PR #1, opened 2024-02-08
Added implementation of buffer

JohanMabille:


CMakeLists.txt
```diff
@@ -37,7 +37,8 @@ OPTION(BUILD_TESTS "xparrow test suite" OFF)
 # =====
 
 set(XPARROW_HEADERS
```
jjerphan, line 39: Not related to this PR: currently headers aren't installed because of difference in case.

This line must be change to:
```suggestion
set(Xparrow_HEADERS
```

or `Xparrow_HEADERS` bellow must be changed to `XPARROW_HEADERS`.

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+    void xbuffer<T>::deallocate(pointer mem) const
+    {
+        delete[] mem;
+    }
+
+    template <class T>
+    bool operator==(const xbuffer<T>& lhs, const xbuffer<T>& rhs)
+    {
+        return lhs.equal(rhs);
+    }
```
jjerphan, line 237: This is required by doctest.
```suggestion
    template <class T>
    bool operator==(const xbuffer<T>& lhs, const xbuffer<T>& rhs)
    {
        return lhs.equal(rhs);
    }
    
    template <class T>
    bool operator!=(const xbuffer<T>& lhs, const xbuffer<T>& rhs)
    {
        return !(lhs == rhs);
    }
```

jjerphan: Excuse my interjection, this is not needed anymore with C++20.

jjerphan (approved):
LGTM apart from one observation.

Thank you, @JohanMabille.

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+        using size_type = typename base_type::size_type;
+
+        xbuffer() = default;
+        explicit xbuffer(size_type size);
+        xbuffer(pointer data, size_type size);
+        
+        ~xbuffer();
+        
+        xbuffer(const xbuffer&);
+        xbuffer& operator=(const xbuffer&);
```
Klaim, line 64: same points as previously with assignations

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+
+    template <class T>
+    bool operator==(const xbuffer<T>& lhs, const xbuffer<T>& rhs);
+
+    /**
+     * @class xbuffer_reference
+     * @brief Object that wraps a piece of contiguous memory
+     * but does not own it.
+     */
+    /*class xbuffer_reference : private xbuffer_data
```
Klaim, line 92: Why is this commented? Or should it be removed?

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+        auto xbuffer_data<T>::size() const noexcept -> size_type
+        {
+            return m_size;
+        }
+
+        template <class T>
+        template <class U>
+        U* xbuffer_data<T>::data() noexcept
+        {
+            return reinterpret_cast<U*>(p_data);
```
Klaim, line 128: shouldnt `static_cast` make more sense here? it's not like it's converting types to/from `void*` so both types needs to have some form of compatibility.

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+    xbuffer<T>::xbuffer(pointer data, size_type size)
+        : base_type{data, size}
+    {
+    }
+        
+    template <class T>
+    xbuffer<T>::~xbuffer()
+    {
+        deallocate(this->p_data);
+        this->p_data = nullptr;
```
Klaim, line 172: Arent these 2 lines unnecessary at this point?

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+    namespace impl
+    {
+        template <class T>
+        bool xbuffer_data<T>::empty() const noexcept
+        {
+            return size() == size_type(0);
+        }
+
+        template <class T>
+        auto xbuffer_data<T>::size() const noexcept -> size_type
```
Klaim, line 119: Do we really need both function and member `size()` knowing that it's an implementation detail? Are we using some algorithm that needs the existence of `size()`? (oh BTW C++23 have `std::size(x);` now, for info)

test/test_xbuffer.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    TEST_SUITE("xbuffer")
+    {
+        TEST_CASE("constructors")
+        {
+            buffer_test_type b0;
+            CHECK_EQ(b0.data(), nullptr);
+            CHECK_EQ(b0.size(), 0u);
+
+            std::size_t size = 4u;
```
Klaim, line 37: `const` and consider renaming to `expected_size` to clarify it's role here.

test/test_xbuffer.cpp
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        TEST_CASE("move semantic")
+        {
+            std::size_t size = 4;
+            buffer_test_type b1(make_test_buffer(size), size);
+            buffer_test_type control(b1);
+            buffer_test_type b2(std::move(b1));
+            CHECK_EQ(b2, control);
+            CHECK_EQ(b1.size(), 0);
```
Klaim, line 70: Note that if `b2` was not empty, then at this point `b1` would not be empty. I'm thinking we might want to not swap in the move operations and release immediately  (on move) the moved-from object internals. Any reason against?

test/test_xbuffer.cpp
```diff
@@ -0,0 +1,82 @@
...
+            CHECK_EQ(b1.data(), nullptr);
+
+            std::size_t size2 = 8;
+            buffer_test_type b4(make_test_buffer(size2, 4), size2);
+            buffer_test_type control2(b4);
+            b2 = std::move(b4);
+            CHECK_EQ(b2, control2);
+            CHECK_EQ(b4, control);
+        }
+    }
```
Klaim, line 81: Please add tests for:
- equality comparison
- swaping
- a super basic test where the object is created and destroyed immediately
- another basic test wher the object is created, filled with values, then destroyed (these 2 are not equivalent to the constructor tests above)
- resizing
- basic checks (maybe in existing tests) that data written in a buffer is the data obtained by reading (in particular after moves and copies would be helpful in the future)

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+        struct xbuffer_data
+        {
+            using value_type = T;
+            using pointer = T*;
+            using size_type = std::size_t;
+            
+            bool empty() const noexcept;
+            size_type size() const noexcept;
+
+            template <class U = T>
```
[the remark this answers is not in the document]
JohanMabille: Because we may need to have some buffer of "raw" memory (i.e. `uint8_t`) that we need to reinterpret as another type.

include/xparrow/xbuffer.hpp
```diff
@@ -0,0 +1,239 @@
...
+    {
+    public:
+
+        using base_type = impl::xbuffer_data<T>;
+        using value_type = typename base_type::value_type;
+        using pointer = typename base_type::pointer;
+        using size_type = typename base_type::size_type;
+
+        xbuffer() = default;
+        explicit xbuffer(size_type size);
```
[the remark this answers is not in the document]
JohanMabille: This constructors allocates memory, so I'm not sure we can use `noexcept` here.

include/xparrow/xbuffer.hpp
JohanMabille: It should be added later actually. That' a leftover that I can remove.

include/xparrow/xbuffer.hpp
JohanMabille: We might convert from / to `uint8_t*` if we store buffer of raw memory.

include/xparrow/xbuffer.hpp
JohanMabille: For the moment I agree that they are not mandatory.

include/xparrow/xbuffer.hpp
JohanMabille: This is exposed in the inheriting class as a public API. Later, `xbuffer_reference` will do the same, so this avoids code duplicatin.

test/test_xbuffer.cpp
JohanMabille: I find the implementation with `swap` quite elegant, otherwise we would have to:
- either deallocating the buffer in the destination, assign the pointer from source, reset the pointer in source
- or swap, and then immediatly delete in the source.

That's more code for "almost" the same result (we don't really care that the memory is not freed immediatly).

The test is misleading here, the only think I want to check is that `b1` did not keep a copy of the pointer, leading to a double destruction. And I want to check this with an fialure rather than a crash (only).

include/xparrow/xbuffer.hpp
Klaim: I would prefer a specific operator for that then, something like `raw_data()` or `bytes()`. Cause right now it's allowed to convert to anything without any warning.

include/xparrow/xbuffer.hpp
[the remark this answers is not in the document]
Klaim: Se my comment in the implementation.

include/xparrow/xbuffer.hpp
JohanMabille: That would handle the case for converting *to* `uint8_t`, but we may also want to convert *from* `uint8_t`, to any unrelated type.

Klaim: I dont think I understand what you mean here?

test/test_xbuffer.cpp
Klaim: I just worry that having data in the moved-from object will lead to false-positive situations when reading them.

Well as long as the size says zero I guess it's ok, accessing the values is undefined-behavior (we might want to specify the contract although it's kind of obvious).

include/xparrow/xbuffer.hpp
JohanMabille: we may use `xbuffer<uint8_t>` as a storage for raw memory. Then, in an array class that depends on it, we know (from somewhere) that's the actual data type is `double`, so we may want to do something like:
```cpp
double* dagta_buffer = buffer.data<double>();
```

The actual type could be any integral type, or floating point, otr whatever else.

Klaim: I am not convinced that this is useful as it hides a potentially dangerous cast in an apparently safe call.
I'm not against going with it and seeing where it leads, but my gut feeling is that it will hide bugs.

test/test_xbuffer.cpp
```diff
@@ -0,0 +1,140 @@
...
+            buffer_test_type b3(std::move(b1));
+            CHECK_EQ(b3.data()[idx], expected_value);
+        }
+
+        TEST_CASE("equality comparison")
+        {
+            const std::size_t size = 4;
+            buffer_test_type b1(make_test_buffer(size), size);
+            buffer_test_type b2(make_test_buffer(size), size);
+            CHECK(b1 == b2);
```
Klaim, line 111: Also `!=`

Klaim (approved):
We can do a `noexcept` pass later, but it needs to be done at some point.

Other than my remarks about seeing if it works, LGTM

test/test_xbuffer.cpp
JohanMabille: :sweat_smile:

PR #4, opened 2024-02-13
Use constexpr constants instead of defines.

Klaim:
This is both to avoid using the preprocessor in C++ in general, and to prepare for future usage of the library as a C++ module.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,9 +28,9 @@ set(SPARROW_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include)
 # ===========
 
 file(STRINGS "${SPARROW_INCLUDE_DIR}/sparrow/sparrow_version.hpp" sparrow_version_defines
-     REGEX "#define SPARROW_VERSION_(MAJOR|MINOR|PATCH)")
+     REGEX "constexpr int SPARROW_VERSION_(MAJOR|MINOR|PATCH)")
 foreach(ver ${sparrow_version_defines})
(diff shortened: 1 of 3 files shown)


include/xparrow/xparrow_version.hpp
```diff
@@ -5,11 +5,11 @@
...
-#define XPARROW_VERSION_MAJOR 0
-#define XPARROW_VERSION_MINOR 0
-#define XPARROW_VERSION_PATCH 1
-
-#endif
+namespace xparrow
+{
+    constexpr int XPARROW_VERSION_MAJOR = 0;
+    constexpr int XPARROW_VERSION_MINOR = 0;
+    constexpr int XPARROW_VERSION_PATCH = 1;
```
JohanMabille, line 14: You need to update the [CMakeLists.txt](https://github.com/xtensor-stack/xparrow/blob/main/CMakeLists.txt#L18) accordingly

Klaim: Should be fixed now, but I didnt rename yet

jjerphan (commented):
How about using [`constinit`](https://en.cppreference.com/w/cpp/language/constinit)?

Klaim:
> How about using [`constinit`](https://en.cppreference.com/w/cpp/language/constinit)?

`constinit` is useful for *runtime* static/global/namespace objects to guarantee that their *initialization* is compile-time, but they are neither `const` nor compile-time themselves.

The version number needs to be both compile-time and runtime and be kept `const` (bascially immutable), so the only option is `constexpr`.

`constinit` can be confusing, it's basically only useful to help with the "static init fiasco" issue.

jjerphan (approved):

JohanMabille (approved):

PR #9, opened 2024-02-29
Added buffer_view class

JohanMabille:


include/sparrow/buffer.hpp
```diff
@@ -90,6 +90,34 @@ namespace sparrow
...
+
+        using base_type = impl::buffer_data<T>;
+        using value_type = typename base_type::value_type;
+        using pointer = typename base_type::pointer;
+        using size_type = typename base_type::size_type;
+
+        explicit buffer_view(buffer<T>& buffer);
+        buffer_view(pointer data, size_type size);
+
+        using base_type::empty;
```
jjerphan, line 110: This is not covered by the tests.

Should it be?

jjerphan (approved):
LGTM.

I just have one comment.

include/sparrow/buffer.hpp
JohanMabille: Yep it should, will update the PR.

PR #11, opened 2024-03-06
Use recent gcc/apple-clang with OSX

Klaim:


Alex-PLACET (approved):

jjerphan:
I do not think we need to use gcc with macOS: this is not a requirement for us, at least now.

Also the latest version of Apple Clang seems to be bugued (see 2576371).

Klaim:
> I do not think we need to use gcc with macOS: this is not a requirement for us, at least now.
> 


I named "gcc/apple-clang" because if you try to use the default gcc  on OSX (as far as I know) it's actually an alias (or symlink) to apple-clang by default. Thats why this pr didnt try to add gcc.

PR #13, opened 2024-03-08
Added reference_proxy and bitset_view

JohanMabille:


include/sparrow/array_data.hpp
```diff
@@ -0,0 +1,279 @@
...
+        static constexpr std::size_t s_bits_per_block = sizeof(block_type) * CHAR_BIT;
+    };
+
+    struct array_data
+    {
+        data_descriptor type;
+        std::int64_t length = 0;
+        std::int64_t null_count = 0;
+        std::int64_t offset = 0;
+        std::vector<buffer<std::uint8_t>> buffers;
```
jjerphan, line 58: Should we define and use?

```cpp
using block_type = std::uint8_t;
```

include/sparrow/data_type.hpp
```diff
@@ -0,0 +1,60 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+namespace sparrow
+{
+    enum class data_type
```
jjerphan, line 19: Should we mention that not [all types](https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings) are supported, yet?

jjerphan (commented):
A first pass.

include/sparrow/array_data.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+#include <climits>
+#include <cstdint>
+#include <vector>
+
+#include "sparrow/buffer.hpp"
+#include "sparrow/dynamic_bitset.hpp"
+#include "sparrow/data_type.hpp"
+
+namespace sparrow
```
Klaim, line 24: Could you add some documentation for each one of these types?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -0,0 +1,338 @@
...
+
+        static constexpr std::size_t s_bits_per_block = sizeof(block_type) * CHAR_BIT;
+
+        storage_type m_buffer;
+        size_type m_size;
+        size_type m_null_count;
+    };
+
+    template <class T>
+    class dynamic_bitset : public dynamic_bitset_base<buffer<T>>
```
Klaim, line 79: These types needs documentation :) Also the constructors because of the differences of ownership.

Question: as-is, `dynamic_bitset<T>` is *not* a range (begin/end/empty are missing). Is this on purpose?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -0,0 +1,338 @@
...
+namespace sparrow
+{
+    template <class B>
+    class dynamic_bitset_base
+    {
+    public:
+
+        using self_type = dynamic_bitset_base<B>;
+        using storage_type = B;
+        using block_type = typename storage_type::value_type;
```
Klaim, line 30: Looks like there are some requirements on `B`. Maybe that is a sign for the need of a concept?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -0,0 +1,338 @@
...
+        using value_type = bool;
+        using size_type = typename storage_type::size_type;
+
+        size_type size() const noexcept;
+        size_type null_count() const noexcept;
+
+        bool test(size_type pos) const;
```
Klaim, line 37: `test` and `set` are not `noexcept`. Is that on purpose?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -0,0 +1,338 @@
...
+
+    template <class B>
+    auto dynamic_bitset_base<B>::null_count() const noexcept -> size_type
+    {
+        return m_null_count;
+    }
+
+    template <class B>
+    bool dynamic_bitset_base<B>::test(size_type pos) const
+    {
```
Klaim, line 143: Consider adding assertion on all functions that could fail because of incorrect input.

include/sparrow/dynamic_bitset.hpp
```diff
@@ -0,0 +1,338 @@
...
+        return !m_null_count || m_buffer.data()[block_index(pos)] & bit_mask(pos);
+    }
+    
+    template <class B>
+    void dynamic_bitset_base<B>::set(size_type pos, value_type value)
+    {
+        block_type& block = m_buffer.data()[block_index(pos)];
+        bool old_value = block & bit_mask(pos);
```
Klaim, line 151: `const`

include/sparrow/dynamic_bitset.hpp
```diff
@@ -0,0 +1,338 @@
...
+    {
+        return bits_count / s_bits_per_block
+            + static_cast<size_type>(bits_count % s_bits_per_block != 0);
+    }
+
+    template <class B>
+    auto dynamic_bitset_base<B>::count_non_null() const noexcept -> size_type
+    {
+        // Number of bits set to 1 in i for i from 0 to 255
+        static constexpr unsigned char table[] =
```
Klaim, line 227: This needs documentation 🥲

include/sparrow/array_data.hpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+        data_descriptor type;
+        std::int64_t length = 0;
+        std::int64_t offset = 0;
+        using block_type = std::uint8_t;
+        using bitmap_type = dynamic_bitset<block_type>; 
+        // bitmap buffer and null_count
+        bitmap_type bitmap;
+        // Other buffers
+        std::vector<buffer<block_type>> buffers;
```
Klaim, line 36: Why define `bitmap_type` but not `buffer_type`? With or without aliases are both fine to me, although I would prefer if it was consistent.

include/sparrow/buffer.hpp
```diff
@@ -26,12 +26,23 @@ namespace sparrow
...
 
+            reference operator[](size_type);
+            const_reference operator[](size_type) const;
+
+            reference front();
+            const_reference front() const;
+
+            reference back();
+            const_reference back() const;
+
```
Klaim, line 45: How about `begin` and `end`? Is this not a range/container?

include/sparrow/buffer.hpp
```diff
@@ -226,11 +289,29 @@ namespace sparrow
         if (n != size())
         {
             buffer<T> tmp(n);
-            std::copy(data(), data() + size(), tmp.data());
+            size_type copy_size = std::min(size(), n);
```
Klaim, line 292: Is this for some exception safety?

include/sparrow/buffer.hpp
```diff
@@ -226,11 +289,29 @@ namespace sparrow
...
+            std::copy(data(), data() + copy_size, tmp.data());
             swap(tmp);
         }
     }
 
+    template <class T>
+    void buffer<T>::resize(size_type n, value_type value)
+    {
+        size_type old_size = size();
```
Klaim, line 301: `const`

include/sparrow/data_type.hpp
```diff
@@ -0,0 +1,62 @@
...
+        FIXED_SIZE_BINARY
+    };
+
+    // For now, a tiny wrapper around data_type
+    // More data and functions to come
+    class data_descriptor
+    {
+    public:
+
+        constexpr explicit data_descriptor(data_type id = data_type::UINT8)
```
Klaim, line 50: I would prefer  a separate default constructor which calls the explicit constructor with 1 parameter. That we we dont have an explicit default constructor.

include/sparrow/dynamic_bitset.hpp
```diff
@@ -0,0 +1,338 @@
...
+        if (extra_bits != 0)
+        {
+            m_buffer.back() &= ~(~block_type(0) << extra_bits);
+        }
+    }
+
+    template <class B>
+    void dynamic_bitset_base<B>::resize(size_type n, value_type b)
+    {
+        size_type old_block_count = m_buffer.size();
```
Klaim, line 285: image

include/sparrow/dynamic_bitset.hpp
JohanMabille: Yes, these are the equivalent of `operator[]` of a container like `std::vector`. I did not implement such operators because these would require other reference proxies that we don't really need for now.

include/sparrow/buffer.hpp
JohanMabille: yeah they should be, and I want to refactor this hierarchy to be consistent with the dynamic bitset. We can add it in an incoming PR though, to keep this one quite "small" (it is actually already big)

include/sparrow/buffer.hpp
JohanMabille: This is to handle resizing with a smaller size.

include/sparrow/buffer.hpp
```diff
@@ -174,6 +230,13 @@ namespace sparrow
     {
     }
 
+    template <class T>
+    buffer<T>::buffer(size_type size, value_type value)
+        : base_type{allocate(size), size}
+    {
+        std::fill(data(), data() + size, value);
```
jjerphan, line 237: As an alternative.
```suggestion
        std::fill_n(data(), size, value);
```

include/sparrow/buffer.hpp
```diff
@@ -226,11 +289,29 @@ namespace sparrow
...
     }
 
+    template <class T>
+    void buffer<T>::resize(size_type n, value_type value)
+    {
+        size_type old_size = size();
+        resize(n);
+        if (old_size < n)
+        {
+            std::fill(data() + old_size, data() + n, value);
```
jjerphan, line 305: Alternatively, one can use:

```suggestion
            std::fill_n(data() + old_size, n, value);
```

test/test_buffer.cpp
```diff
@@ -156,6 +165,23 @@ namespace sparrow
             b.resize(size2);
             CHECK_EQ(b.size(), size2);
             CHECK_EQ(b.data()[2], 2);
+
+            b.resize(size1);
+            CHECK_EQ(b.size(), size1);
+            CHECK_EQ(b.data()[2], 2);
+
+            const std::size_t size3 = 6u;
```
jjerphan, line 173: Should we move it above?

include/sparrow/array_data.hpp
```diff
@@ -0,0 +1,279 @@
...
+        template <class U>
+        self_type& operator-=(const U&);
+
+        template <class U>
+        self_type& operator*=(const U&);
+
+        template <class U>
+        self_type& operator/=(const U&);
+
+        operator const_reference() const;
```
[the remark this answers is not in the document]
Klaim: > How having this operator implicit can be harmful?

See [this guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c164-avoid-implicit-conversion-operators) for the fundations of why it's a *very bad idea*. It goes in pair with the guideline about 

> This requires that you explicitly cast each time you want to compare the proxy to a value (for instance), which is paricularly ugly (and does not looks naturaly at all to me):

That just means that the reference type needs comparison operators with objects of the type being referred to. That conversion is a shortctut that can cause more issues than explicitly stating these operators.

Also note that even like that "ugly" is not a reason to make the code potentially silently incorrect, in particular in a C++ library supposedly trying to make correctness easier than the C api. Worst scenario, `static_cast`s are fine and help clarify where are the potential issues related to conversion. They are noisy for a reason after all.

See as example: https://godbolt.org/z/7jxcvjq3Y

I see no reason to make that operator implicit.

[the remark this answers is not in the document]
Klaim: I forgot a test in my example, here is the udpated version (I only added the test): https://godbolt.org/z/TYx7sGezj

test/test_dynamic_bitset.cpp
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            const std::size_t expected_block_count = 2;
+            std::uint8_t* buf = new std::uint8_t[expected_block_count];
+            buf[0] = 37;
+            buf[1] = 2;
+            bitmap b3(buf, expected_block_count * 8);
+            
+            b2 = b3;
+            CHECK_EQ(b2.size() , b3.size());
```
jjerphan, line 101:
```suggestion
            CHECK_EQ(b2.size(), b3.size());
```

include/sparrow/buffer.hpp
Klaim: Consider adding TODOs for these so that we dont forget in further passes.

include/sparrow/dynamic_bitset.hpp
jjerphan: Isn't the comment sufficient?

jjerphan (commented):
Another pass.

JohanMabille:
Marking it as draft to avoid accidental merge. This PR will be split into many smaller ones.

jjerphan:
Can this PR be closed now?

PR #16, opened 2024-03-11
Refactored buffer class and added access methods

JohanMabille:
Partially replaces #13

### DIFF
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -19,8 +19,63 @@
 #include <concepts>
 #include <cstdint>
 
+#include "sparrow/iterator.hpp"
+
 namespace sparrow
 {
+    namespace impl
+    {
+        // Should probably move to a MP util file
+        template <class T, bool is_const>
+        struct constify
+            : std::conditional<is_const, const T, T>
+        {
+        };
+
+        template <class T, bool is_const>
+        using constify_t = typename constify<T, is_const>::type;
+    }
+
+    template <class T, bool is_const>
+    class buffer_iterator
+        : public iterator_base
+          <
+              buffer_iterator<T, is_const>,
+              impl::constify_t<T, is_const>,
+              std::contiguous_iterator_tag
+          >
+    {
+    public:
+
+        using self_type = buffer_iterator<T, is_const>;
+        using base_type = iterator_base
+        <
+            self_type,
+            impl::constify_t<T, is_const>,
+            std::contiguous_iterator_tag
+        >;
+        using pointer = typename base_type::pointer;
+
+        explicit buffer_iterator(pointer p = nullptr);
+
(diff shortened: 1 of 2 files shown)


include/sparrow/buffer.hpp
```diff
@@ -15,53 +15,172 @@
...
-        struct buffer_data
+        // Should probably move to a MP util file
+        template <class T, bool is_const>
+        struct constify
+            : std::conditional<is_const, const T, T>
         {
-            using value_type = T;
-            using pointer = T*;
-            using size_type = std::size_t;
+        };
```
jjerphan, line 33: Would you use it for all the implementations?

jjerphan (approved):
I just have a comment and a nitpick.

include/sparrow/buffer.hpp
JohanMabille: I don't know yet for sure, thus the "probably" in the comment above.

test/test_buffer.cpp
```diff
@@ -156,6 +197,53 @@ namespace sparrow
             b.resize(size2);
             CHECK_EQ(b.size(), size2);
             CHECK_EQ(b.data()[2], 2);
+
+            b.resize(size1);
+            CHECK_EQ(b.size(), size1);
+            CHECK_EQ(b.data()[2], 2);
+
+            const std::size_t size3 = 6u;
```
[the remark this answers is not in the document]
JohanMabille: I prefer to define variables the latest possible, just before they're used.

include/sparrow/buffer.hpp
```diff
@@ -15,53 +15,172 @@
...
 
+#include "sparrow/iterator.hpp"
+
 namespace sparrow
 {
     namespace impl
     {
-        template <class T>
-        struct buffer_data
+        // Should probably move to a MP util file
```
Klaim, line 28: Consider adding `TODO` so that we can easilly search this kidn of thing

Klaim (approved):
Mostly needs a bit more documentation but LGTM otherwise.

include/sparrow/buffer.hpp
```diff
@@ -15,53 +15,172 @@
...
+        using self_type = buffer_iterator<T, is_const>;
+        using base_type = iterator_base
+        <
+            self_type,
+            impl::constify_t<T, is_const>,
+            std::contiguous_iterator_tag
+        >;
+        using pointer = typename base_type::pointer;
+
+        explicit buffer_iterator(pointer p = nullptr);
```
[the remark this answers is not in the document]
JohanMabille: > If `p == nullptr`, does this makes the iterator instance a sentinel?

Actually it's a remainder of Legacy iterators, which had to be default constructible. Let's remove the édefault" constructor.

PR #19, opened 2024-03-13
Move `constify_t` under metaprog utilities

jjerphan:


include/sparrow/mp_utils.hpp
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <iterator>
+
+namespace sparrow
+{
+    namespace impl
```
JohanMabille, line 21: NItipicking: maybe we can take the opportunity to rename this namespace mpl

JohanMabille (approved):

PR #20, opened 2024-03-13
Added contraints to dynamic_bitset types

JohanMabille:
Small PR to add constraints to dynamic bitset types.

### DIFF
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -56,6 +56,8 @@ namespace sparrow
         >;
         using pointer = typename base_type::pointer;
 
+        // Required to be used seen as a range iterator
+        buffer_iterator() = default;
         explicit buffer_iterator(pointer p);
 
     private:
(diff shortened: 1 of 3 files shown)


include/sparrow/buffer.hpp
```diff
@@ -56,6 +56,8 @@ namespace sparrow
         >;
         using pointer = typename base_type::pointer;
 
+        // Required to be used seen as a range iterator
```
jjerphan, line 59: What is a "range iterator" in this case?

Is the following what you mean?
```suggestion
        // Required to be used seen as a `random_access_iterator`
```

jjerphan (approved):
LGTM.

include/sparrow/buffer.hpp
JohanMabille: no I meant that `std::ranges::end(b)` where `b` is a buffer is valid.

PR #21, opened 2024-03-13
Add project version

jjerphan:


JohanMabille:
The version of the library is already exposed in [sparrow_version.hpp](https://github.com/xtensor-stack/sparrow/blob/main/include/sparrow/sparrow_version.hpp), I am not sure that we want to maintain it at two different places. I would be in favor of keeping the C++ one rather than the cmke one.

Issue #14, opened 2024-03-11
Implement basic layouts


JohanMabille:
The first layouts to implement must be
- [primitive_layout](https://arrow.apache.org/docs/dev/format/Columnar.html#fixed-size-primitive-layout)
- [binary_layout](https://arrow.apache.org/docs/dev/format/Columnar.html#variable-size-binary-layout)
- [dictionary_layout](https://arrow.apache.org/docs/dev/format/Columnar.html#dictionary-encoded-layout)

The different layouts should provide the following API:

```cpp
template <class T>
struct arbitrary_layout
{
    class iterator : public iterator_base<iterator>
    {
        // ...
    };

    class const_iterator : public iterator_base<const_iterator>
    {
        // ...
    };

    reference element(size_type i, array_data& data);
    const_reference element(size_type i, const array_data& data) const;

    iterator begin(array_data& data);
    iterator end(array_data& data);

    const_iterator begin(array_data& data) const;
    const_iterator end(array_data& data) const;
};
```

NB:
- the methods otf the layouts could actually be static since there is not state in the class iteself. Not sure which solution is better
- we will probably figure out possible factorization when we have many layouts implemented.

JohanMabille:
The three basic layouts have been implemented in the PR above.

Issue #15, opened 2024-03-11
Implement metaprogramming utilities


JohanMabille:
This issue is too big to be implmeented in a single PR and has been split into smaller issues (the list will be updated while doing the work):

- [X] Metaprogramming basic tools: https://github.com/xtensor-stack/sparrow/issues/67
- [X] Support for half float type: https://github.com/xtensor-stack/sparrow/issues/68

The target architecture is something like:

```cpp
struct array_data
{
        data_descriptor type;
        std::int64_t length = 0;
        std::int64_t offset = 0;
        dynamic_bitset<std::uint8_t> bitmap;       // buffer 0 in C struct
        std::vector<buffer<std::uint8_t>> buffers; // buffer 1 to n in C struct 
        std::vector<array_data> child_data;
};

template <class T, class L = get_default_layout_t<T>>
class array_impl
{
    L m_layout;
    array_data m_data;
};

using array_variant = std::variant<
    array_impl<std::unint8_t>,
    array_impl<std::int8_t>,
    ...
    // all required types already supported in data_type.hpp
>;
```

We need to implement tools such as arrow_traits to retrieve the different associated types to a basic type:

```cpp
template <class T>
struct arrow_traits;

template <>
struct arrow_traits<std::unint8_t>
{
    using layout_type = primite_layout<std::uint8_t>;
};
```

It would be nice to use concepts / constraints to reduce the amount of code, and to be able to generate `array_impl` and other related types based on a typelist of supported types.

JohanMabille:
This issue is too big to be implemented in a single PR; see below the list of smallest issues for implementing it (will be updated while doing the work):

- Metaprogramming basic tools: #67 
- Support for half float type: #68

Klaim:
@JohanMabille Is there something missing or should we close this?

JohanMabille:
We can close it.

PR #22, opened 2024-03-13
Primitive layout

jjerphan:
Comes after #13.

Partially addresses #14.

### DIFF
--- a/include/sparrow/layout.hpp
+++ b/include/sparrow/layout.hpp
@@ -25,9 +25,9 @@
 namespace sparrow
 {
     /**
-     * An iterator for primitive_layout.
+     * An iterator for `primitive_layout` operating on contiguous data only.
      *
-     * @tparam T The type of the elements in the layout.
+     * @tparam T The type of the elements in the layout's data buffer.
      * @tparam is_const A boolean indicating whether the iterator is const.
      *
      * @note This class is not thread-safe, exception-safe, copyable, movable, equality comparable.
@@ -52,8 +52,6 @@ namespace sparrow
         >;
         using pointer = typename base_type::pointer;
 
-        // Required so that std::ranges::end(p) is
-        // valid when p is a primitive_layout.
         primitive_layout_iterator() = default;
         explicit primitive_layout_iterator(pointer p);
 
@@ -72,7 +70,6 @@ namespace sparrow
         bool equal(const self_type& rhs) const;
         bool less_than(const self_type& rhs) const;
 
-        // We only use the first buffer and the bitmap.
         pointer m_pointer = nullptr;
 
         friend class iterator_access;
@@ -85,7 +82,8 @@ namespace sparrow
      * It iterates over the first buffer in the array_data, and uses the bitmap to skip over null.
      * The bitmap is assumed to be present in the array_data.
      *
-     * @tparam T The type of the elements in the layout.
+     * @tparam T The type of the elements in the layout's data buffer.
+     *           A primitive type or a fixed size type.
      *
      * @note This class is not thread-safe, exception-safe, copyable, movable, equality comparable.
      */
@@ -220,13 +218,13 @@ namespace sparrow
     template <class T>
     auto primitive_layout<T>::begin() -> iterator
     {
-        return iterator(data());
+        return iterator{data()};
     }
 
     template <class T>
     auto primitive_layout<T>::end() -> iterator
     {
-        return iterator(data()) + self_type::size();
+        return iterator{data() + self_type::size()};
     }
 
     template <class T>
@@ -244,13 +242,13 @@ namespace sparrow
     template <class T>
     auto primitive_layout<T>::cbegin() const -> const_iterator
     {
-        return const_iterator(data());
+        return const_iterator{data()};
     }
 
     template <class T>
     auto primitive_layout<T>::cend() const -> const_iterator
     {
-        return const_iterator(data()) + self_type::size();
+        return const_iterator{data() + self_type::size()};
     }
 
     template <class T>
--- a/include/sparrow/layout.hpp
+++ b/include/sparrow/layout.hpp
@@ -256,12 +256,14 @@ namespace sparrow
     template <class T>
     auto primitive_layout<T>::data() -> pointer
     {
+        assert(m_data.buffers.size() > 0);
         return m_data.buffers[0].data();
     }
 
     template <class T>
     auto primitive_layout<T>::data() const -> const_pointer
     {
+        assert(m_data.buffers.size() > 0);
         return m_data.buffers[0].data();
     }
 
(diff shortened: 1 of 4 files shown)


test/test_layout.cpp
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "doctest/doctest.h"
+
+#include "sparrow/layout.hpp"
+
+namespace sparrow
+{
+
```
Klaim, line 21: Consider adding `static_assert(std::range< primitive_layout<int>>);` and other compile-time checks on the expected interfaces. Maybe iterators etc.

test/test_layout.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <iostream>
+#include <numeric>
+
+#include "doctest/doctest.h"
+
+#include "sparrow/layout.hpp"
+
+namespace sparrow
+{
+    static_assert(std::ranges::range<primitive_layout<uint8_t>>);
```
Klaim, line 23: Consider adding a `TODO` to test all the other base types (I'm working on maknig this easy, so that would have to wait for #15)

include/sparrow/layout.hpp
```diff
@@ -0,0 +1,251 @@
...
+#include <cstdint>
+
+#include "sparrow/array_data.hpp"
+#include "sparrow/iterator.hpp"
+#include "sparrow/buffer.hpp"
+
+namespace sparrow
+{
+    template <class T, bool is_const>
+    class primitive_layout_iterator
```
Klaim, line 28: Please add some documentation each type. :)

include/sparrow/layout.hpp
```diff
@@ -0,0 +1,251 @@
...
+        <
+            self_type,
+            mpl::constify_t<T, is_const>,
+            std::contiguous_iterator_tag
+        >;
+        using pointer = typename base_type::pointer;
+
+        // Required so that std::ranges::end(p) is
+        // valid when p is a primitive_layout.
+        primitive_layout_iterator() = default;
```
Klaim, line 49: Yes, you might want to add a static check for the kind of iterator it is, see the concepts there: https://en.cppreference.com/w/cpp/iterator (in the "Iterator concepts" section).
Note: the check cannot be inside the type as it is not defined yet, and it cannot be made for all the types in template parameters, except if you write the check for example in the constructor, and the constructor defined outside the class definition. That's a trick for checking constraints on templates being well-formed once instantiated.

include/sparrow/layout.hpp
```diff
@@ -0,0 +1,251 @@
...
+        void increment();
+        void decrement();
+        void advance(difference_type n);
+        difference_type distance_to(const self_type& rhs) const;
+        bool equal(const self_type& rhs) const;
+        bool less_than(const self_type& rhs) const;
+
+        // We only use the first buffer and the bitmap.
+        pointer m_pointer = nullptr;
+        size_type m_index = 0;
```
Klaim, line 69: If I understand correctly, `m_pointer`  is the currently pointed element, but then we have an index instead of just changing `m_pointer`. Why this choice? Why not just iterate the pointer?

include/sparrow/layout.hpp
```diff
@@ -0,0 +1,251 @@
...
+    template <class T>
+    auto primitive_layout<T>::cend() const -> const_iterator
+    {
+        return const_iterator(data()) + self_type::size();
+    }
+
+    template <class T>
+    auto primitive_layout<T>::data() -> pointer
+    {
+        return m_data.buffers[0].data();
```
Klaim, line 240: This assumes that `buffers.size() > 0` but my understanding is that it's not always the case?
@JohanMabille You know best what's required here exactly I think.

JohanMabille: for primitive layout, it is expected that `m_data.buffers.size() == 1`, we should add an assertion in the constructor.

Klaim: The assertion shoudl be here too. There is a situation where the buffer becomes empty: when we pass the ownership to something else.

include/sparrow/layout.hpp
jjerphan: Indeed, before an alternative approach was used, yet `m_index` can be removed.

include/sparrow/layout.hpp
jjerphan: For now an exception is thrown in the constructor, but should we instead just use assertions?

include/sparrow/layout.hpp
jjerphan: Isn't the tag l.33 sufficient?

include/sparrow/layout.hpp
JohanMabille: Yes

include/sparrow/layout.hpp
JohanMabille: The tag is for specifying the iterator type, the `static_assert` is for ensuring that the code is correct and detect whenever we break the base class.

include/sparrow/layout.hpp
```diff
@@ -0,0 +1,272 @@
...
+    template <class T>
+    auto primitive_layout<T>::begin() -> iterator
+    {
+        return iterator(data());
+    }
+
+    template <class T>
+    auto primitive_layout<T>::end() -> iterator
+    {
+        return iterator(data()) + self_type::size();
```
Klaim, line 229: Not important here but general note in case you didnt know: except for standard containers, you can use `{ ... }` to call constructors, in which case implicit conversions are disabled.

test/test_layout.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <numeric>
+
+#include "doctest/doctest.h"
+
+#include "sparrow/layout.hpp"
+
+namespace sparrow
+{
+    // TODO: Test all the other base types once #15 is addressed.
+    static_assert(std::ranges::range<primitive_layout<uint8_t>>);
```
Klaim, line 24: Nice! It just occured to me that in this specific case (primitive_layout) we can test it's a [contiguous_range](https://en.cppreference.com/w/cpp/ranges/contiguous_range) more specifically.

test/test_layout.cpp
```diff
@@ -0,0 +1,85 @@
...
+            ad.buffers.push_back(b);
+            ad.length = n;
+            ad.offset = 0;
+            ad.child_data.push_back(array_data());
+            return ad;
+        }
+
+    }
+
+    TEST_SUITE("layout")
```
Klaim, line 51: Arent these tests specific to `primitive_layout`?

include/sparrow/layout.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+        reference dereference() const;
+        void increment();
+        void decrement();
+        void advance(difference_type n);
+        difference_type distance_to(const self_type& rhs) const;
+        bool equal(const self_type& rhs) const;
+        bool less_than(const self_type& rhs) const;
+
+        // We only use the first buffer and the bitmap.
```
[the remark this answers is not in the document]
jjerphan: Yes, that's exact.

JohanMabille (approved):

PR #24, opened 2024-03-15
Added reference_proxy implementation

JohanMabille:
Last piece missing from  #13

### DIFF
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -53,8 +53,8 @@ namespace sparrow
     /**
      * Reference type used in layout classes.
      * This classes provides a similar behavior
-     * as std::optional, but on references in
-     * an array_data.
+     * as std::optional, but as reference on data
+     * in an array_data.
      */
     template <class L>
     class reference_proxy
@@ -87,10 +87,11 @@ namespace sparrow
         template <std::convertible_to<typename L::inner_value_type> U>
         self_type& operator=(U&& value);
 
-        void reset();
         void swap(self_type& rhs);
         
     private:
+
+        void reset();
 
         template <class U>
         void update(U&& u);
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -211,8 +211,8 @@ namespace sparrow
     template <class L>
     void reference_proxy<L>::swap(self_type& rhs)
     {
-        bool this_has_value = has_value();
-        bool rhs_has_value = rhs.has_value();
+        const bool this_has_value = has_value();
+        const bool rhs_has_value = rhs.has_value();
         using std::swap;
         if (this_has_value)
         {
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -52,7 +52,7 @@ namespace sparrow
 
     /**
      * Reference type used in layout classes.
-     * This classes provides a similar behavior
+     * This class provides a similar behavior
      * as std::optional, but as reference on data
      * in an array_data.
      */
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -14,12 +14,26 @@
 
 #pragma once
 
+#include <cassert>
+#include <compare>
+#include <concepts>
+#include <optional>
+#include <vector>
+
 #include "sparrow/buffer.hpp"
 #include "sparrow/data_type.hpp"
 #include "sparrow/dynamic_bitset.hpp"
 
 namespace sparrow
 {
+    /**
+     * Structure holding the raw data.
+     *
+     * array_data is meant to be used by the
+     * different layout classes to implement
+     * the array API, based on the type
+     * specified in the type attribute.
+     */
     struct array_data
     {
         using block_type = std::uint8_t;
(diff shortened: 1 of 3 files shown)


include/sparrow/array_data.hpp
```diff
@@ -36,8 +50,251 @@ namespace sparrow
         std::vector<array_data> child_data;
     };
 
-    struct null_type
+    /**
+     * Reference type used in layout classes.
+     * This classes provides a similar behavior
+     * as std::optional, but on references in
```
Klaim, line 56: Suggestion: `but as reference on data in an array_data`

test/test_array_data.cpp
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+    private:
+
+        bool has_value(size_type index) const
+        {
+            return m_storage[index].has_value();
+        }
+
+        typename value_type::value_type& value(size_type index)
```
jjerphan, line 55: How about?
```suggestion
        typename inner_value_type& value(size_type index)
```

test/test_array_data.cpp
```diff
@@ -0,0 +1,221 @@
...
+        {
+            return m_storage[index].value();
+        }
+
+        const value_type::value_type& value(size_type index) const
```
jjerphan, line 60:
```suggestion
        const inner_value_type& value(size_type index) const
```

include/sparrow/array_data.hpp
```diff
@@ -36,8 +50,251 @@ namespace sparrow
...
+
+        template <std::convertible_to<typename L::inner_value_type> U>
+        self_type& operator=(std::optional<U>&& rhs);
+
+        template <std::convertible_to<typename L::inner_value_type> U>
+        self_type& operator=(U&& value);
+
+        void reset();
+        void swap(self_type& rhs);
+
```
Klaim, line 92: Consider adding a `// TODO: ...` for adding all the other usual arithmetic operators that might be allowed for `T`?

include/sparrow/array_data.hpp
```diff
@@ -36,8 +50,251 @@ namespace sparrow
...
+        template <std::convertible_to<typename L::inner_value_type> U>
+        self_type& operator=(const std::optional<U>& rhs);
+
+        template <std::convertible_to<typename L::inner_value_type> U>
+        self_type& operator=(std::optional<U>&& rhs);
+
+        template <std::convertible_to<typename L::inner_value_type> U>
+        self_type& operator=(U&& value);
+
+        void reset();
```
Klaim, line 90: It's not clear to me what this does exactly? Needs documentation, like `post-condition: has_value() == false` ?

test/test_array_data.cpp
```diff
@@ -0,0 +1,221 @@
...
+    public:
+
+        using value_type = std::optional<int>;
+        using inner_value_type = value_type::value_type;
+        using reference = reference_proxy<mock_layout>;
+        using size_type = std::size_t;
+        using storage_type = std::vector<std::optional<int>>;
+
+        mock_layout() = default;
+        mock_layout(std::initializer_list<value_type> l)
```
Klaim, line 33: `explicit` :D

test/test_array_data.cpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    class mock_layout
+    {
+    public:
+
+        using value_type = std::optional<int>;
+        using inner_value_type = value_type::value_type;
+        using reference = reference_proxy<mock_layout>;
+        using size_type = std::size_t;
+        using storage_type = std::vector<std::optional<int>>;
```
jjerphan, line 30:
```suggestion
        using storage_type = std::vector<value_type>;
```

Klaim (commented):
Other than missing documentation  on functions and other details, LGTM

test/test_array_data.cpp
```diff
@@ -0,0 +1,221 @@
...
+        {
+            int expected_value0 = stored_value(0);
+            int expected_value3 = stored_value(3);
+
+            auto ref0 = m_layout[0];
+            auto ref2 = m_layout[2];
+            auto ref3 = m_layout[3];
+
+            ref0.swap(ref2);
+        }
```
jjerphan, line 171: This test case looks incomplete. `ref3` is not needed.

test/test_array_data.cpp
```diff
@@ -0,0 +1,221 @@
...
+            auto ref0 = m_layout[0];
+            CHECK_EQ(ref0.value(), stored_value(0));
+            static_assert(std::same_as<decltype(ref0.value()), int&>);
+
+            const auto cref0 = m_layout[0];
+            CHECK_EQ(cref0.value(), stored_value(0));
+            static_assert(std::same_as<decltype(cref0.value()), const int&>);
+        }
+
+        TEST_CASE_FIXTURE(ref_proxy_fixture, "assign")
```
jjerphan, line 123: We might test assigning `std::nullopt_t`.

jjerphan (commented):
The implementation LGTM. I just have comments for the tests.

test/test_array_data.cpp
JohanMabille: Ah indeed, I forgot to replace this. The typename is not even required with the change.

include/sparrow/array_data.hpp
JohanMabille: They don't exist for `std::optional`, that is why I haven't added them.

include/sparrow/array_data.hpp
JohanMabille: I was confused with `reset` from `std::optional`. I'll move this method to the private section.

test/test_array_data.cpp
JohanMabille: not for initializer_list ;)

test/test_array_data.cpp
JohanMabille: It is done line 130

include/sparrow/array_data.hpp
Klaim: Yes, I meant do we think we will need them in the context of arrow db manipulation?

include/sparrow/array_data.hpp
```diff
@@ -36,8 +50,252 @@ namespace sparrow
         std::vector<array_data> child_data;
     };
 
-    struct null_type
+    /**
+     * Reference type used in layout classes.
+     * This classes provides a similar behavior
```
jjerphan, line 55:
```suggestion
     * This class provides a similar behavior
```

jjerphan (approved):

include/sparrow/array_data.hpp
JohanMabille: For now I don't think so. most of the time, people simply perform the operations through the pair of value iterator / bitmask iterator, and perform operations even if the bitmask if false. If your data is not too sparse (which is the case most of the time), it is faster than testing the flag and branching for each value.

PR #25, opened 2024-03-15
Implemented const_reference_proxy and CRTP for reference_proxy

JohanMabille:


test/test_array_data.cpp
```diff
@@ -178,7 +190,7 @@ namespace sparrow
                 auto ref1 = m_layout[1];
                 auto ref2 = m_layout[2];
 
-                CHECK(ref0 != ref1);
+                CHECK(bool(ref0 != ref1));
```
jjerphan, line 193: Why is this required now? Aren't members of `std::strong_ordering` converted to bool?

The same comment applies l.210.

jjerphan (approved):

PR #27, opened 2024-03-18
Implemented bitset_iterator and bitset_reference

JohanMabille:


include/sparrow/dynamic_bitset.hpp
```diff
@@ -228,6 +340,44 @@ namespace sparrow
...
+    template <random_access_range B>
+    auto dynamic_bitset_base<B>::begin() -> iterator
+    {
+        return iterator(this, data(), 0u);
+    }
+
+    template <random_access_range B>
+    auto dynamic_bitset_base<B>::end() -> iterator
+    {
+        block_type* block = m_buffer.size() ? data() + m_buffer.size() - 1 : data();
```
jjerphan, line 352: I guess we need to support empty buffer, yes.

include/sparrow/dynamic_bitset.hpp
```diff
@@ -154,6 +178,101 @@ namespace sparrow
...
+        difference_type distance_to(const self_type& rhs) const;
+        bool equal(const self_type& rhs) const;
+        bool less_than(const self_type& rhs) const;
+
+        bool is_first_bit_of_block(size_type index) const;
+        difference_type distance_to_begin() const;
+
+        bitset_type* p_bitset = nullptr;
+        block_type* p_block = nullptr;
+        size_type m_index;
```
jjerphan, line 271: Could we specify if `m_index` corresponds to the block-local index or the absolute "logical" index?

For instance for the first case:
```suggestion
        // m_index is block-local index.
        // Invariant: 0 <= m_index < bitset_type::s_bits_per_block
        size_type m_index;
```

include/sparrow/dynamic_bitset.hpp
```diff
@@ -380,4 +545,239 @@ namespace sparrow
...
+    template <class B, bool is_const>
+    bool bitset_iterator<B, is_const>::is_first_bit_of_block(size_type index) const
+    {
+        return m_index % bitset_type::s_bits_per_block == 0;
+    }
+
+    template <class B, bool is_const>
+    auto bitset_iterator<B, is_const>::distance_to_begin() const -> difference_type
+    {
+        return bitset_type::s_bits_per_block * (p_block - p_bitset->begin().p_block) + m_index;
```
jjerphan, line 781: I believe this is true iff `0 <= m_index < bitset_type::s_bits_per_block`.

Could we test it?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -154,6 +178,101 @@ namespace sparrow
...
+        self_type& operator=(bool) noexcept;
+
+        explicit operator bool() const noexcept;
+
+        bool operator~() const noexcept;
+
+        self_type& operator&=(bool) noexcept;
+        self_type& operator|=(bool) noexcept;
+        self_type& operator^=(bool) noexcept;
+
```
Klaim, line 205: What about `operator==` and `operator!=`?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -154,6 +178,101 @@ namespace sparrow
         dynamic_bitset_view& operator=(dynamic_bitset_view&&) = default;
     };
 
+    /**
+     *@class bitset_reference
+     */
+    template <class B>
```
Klaim, line 184: What is `B`?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -20,9 +20,16 @@
 #include <ranges>
 
 #include "sparrow/buffer.hpp"
+#include "sparrow/mp_utils.hpp"
 
 namespace sparrow
 {
+    template <class B, bool is_const>
```
Klaim, line 27: These new types needs some documentation :)

include/sparrow/dynamic_bitset.hpp
```diff
@@ -154,6 +178,101 @@ namespace sparrow
...
+
+        void assign(bool) noexcept;
+        void set() noexcept;
+        void reset() noexcept;
+        
+        bitset_type& m_bitset;
+        block_type& m_block;
+        block_type m_mask;
+
+        friend class bitset_iterator<B, false>;
```
Klaim, line 221: I'm not sure if it's worth associating (`friend` being the tightest bound) these types instead of just making the constructor public?

test/test_dynamic_bitset.cpp
```diff
@@ -176,6 +176,77 @@ namespace sparrow
             CHECK_EQ(b.size(), 29);
             CHECK_EQ(b.null_count(), m_null_count);
         }
+
+        TEST_CASE_FIXTURE(bitmap_fixture, "iterator")
```
Klaim, line 180: Add `static_assert`  checking the right concept iterator?

Klaim (commented):
This needs documentation :)

I'll check the branch on windows/msvc+ASAN

include/sparrow/dynamic_bitset.hpp
JohanMabille: we can make assertions when we modify it

include/sparrow/dynamic_bitset.hpp
JohanMabille: The iterator also needs to access some private methods (for computing the blocks indices, index in a block, etc). I could make an bitset_access backdoor to limit the part of the API we expose to the iterator, but I find it a bit of overkill here.

include/sparrow/dynamic_bitset.hpp
```diff
@@ -380,4 +562,249 @@ namespace sparrow
...
+     * bitset_iterator implementation *
+     **********************************/
+
+    template <class B, bool is_const>
+    bitset_iterator<B, is_const>::bitset_iterator(bitset_type* bitset, block_type* block, size_type index)
+        : p_bitset(bitset)
+        , p_block(block)
+        , m_index(index)
+    {
+        assert(m_index < bitset_type::s_bits_per_block);
```
jjerphan, line 681:
```suggestion
        assert(0 <= m_index);
        assert(m_index < bitset_type::s_bits_per_block);
```

include/sparrow/dynamic_bitset.hpp
```diff
@@ -380,4 +562,249 @@ namespace sparrow
...
+    void bitset_iterator<B, is_const>::increment()
+    {
+        ++m_index;
+        // If we have reached next block
+        if (is_first_bit_of_block(m_index))
+        {
+            ++p_block;
+            m_index = 0u;
+        }
+        assert(m_index < bitset_type::s_bits_per_block);
```
jjerphan, line 707:
```suggestion
        assert(0 <= m_index);
        assert(m_index < bitset_type::s_bits_per_block);
```

include/sparrow/dynamic_bitset.hpp
```diff
@@ -380,4 +562,249 @@ namespace sparrow
...
+        if (is_first_bit_of_block(m_index))
+        {
+            --p_block;
+            m_index = sizeof(block_type);
+        }
+        else
+        {
+            --m_index;
+        }
+        assert(m_index < bitset_type::s_bits_per_block);
```
jjerphan, line 723:
```suggestion
        assert(0 <= m_index);
        assert(m_index < bitset_type::s_bits_per_block);
```

include/sparrow/dynamic_bitset.hpp
```diff
@@ -380,4 +562,249 @@ namespace sparrow
...
+                }
+                else
+                {
+                    mn -= m_index;
+                    --p_block;
+                    m_index = bitset_type::s_bits_per_block - mn;
+                }
+            }
+        }
+        assert(m_index < bitset_type::s_bits_per_block);
```
jjerphan, line 769:
```suggestion
        assert(0 <= m_index);
        assert(m_index < bitset_type::s_bits_per_block);
```

jjerphan (approved):

include/sparrow/dynamic_bitset.hpp
Klaim: `m_index` cannot be inferior to zero because it's unsigned. Not sure if there is a max value we could check however.

include/sparrow/dynamic_bitset.hpp
```diff
@@ -380,4 +562,249 @@ namespace sparrow
...
+    {
+        // decreasing moves to previous block
+        if (is_first_bit_of_block(m_index))
+        {
+            --p_block;
+            m_index = sizeof(block_type);
+        }
+        else
+        {
+            --m_index;
```
Klaim, line 721: Is this what we want when `m_index` is zero?
Note that the valeu would then switch to the highest index, and that's *defined behavior*. But is it valid for us? Should we assert here?
Or should `m_index` be changed to a difference type instead (which would then be signed)?

JohanMabille: If `m_index` is 0, then `is_first_bit_of_block(m_index)` is true and we never reach the `else`.

Klaim (approved):

PR #30, opened 2024-03-18
Various Minor CI fixes

Klaim:
- using `ctest` to run the tests, which means we dont have to change the command executing tests every time we want to add a test executable or configuration;
- specifies `Debug` and `Release` configurations to be explicitly built and tested, to avoid issues because of CMake's inconsistency in the output of generators (some build-systems allow multiple configurations, so that removes any ambiguity) - also for installing;
- builds with both Ninja and VS on Windows, with msvc, mingw and clang;
- git-ignore some poentially generated directories
- fixed some ci runs names

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -66,11 +66,11 @@ jobs:
 
     - name: Install
       working-directory: build
-      run: cmake --install .
+      run: cmake --install . --config ${{matrix.config.name}}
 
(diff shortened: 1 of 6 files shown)


Klaim:
I updated the first message here to reflect the intent of this PR.

jjerphan (approved):

JohanMabille (approved):

PR #31, opened 2024-03-18
Adapt primitive_layout into fixed_size_layout

jjerphan:


include/sparrow/layout.hpp
```diff
@@ -94,11 +94,14 @@ namespace sparrow
...
-        using pointer = T*;
-        using const_pointer = const T*;
+        using inner_value_type = T;
+        using value_type = std::optional<inner_value_type>;
+        using inner_reference = inner_value_type&;
+        using inner_const_reference = const inner_reference;
+        using reference = reference_proxy<self_type>;
+        using const_reference = const_reference_proxy<self_type>;
+        using pointer = inner_value_type*;
+        using const_pointer = const inner_value_type*;
```
JohanMabille, line 104: These are the definitions for `inner_pointer` and `inner_const_pointer`. I don't think we need the pointer types for now.

include/sparrow/layout.hpp
```diff
@@ -108,8 +111,10 @@ namespace sparrow
         using const_iterator = primitive_layout_iterator<T, true>;
 
         size_type size() const;
-        reference element(size_type i);
-        const_reference element(size_type i) const;
+        inner_reference value(size_type i);
+        inner_const_reference value(size_type i) const;
+
+        bool has_value(size_type i) const;
```
JohanMabille, line 117: I wonder if the `value` / `has_value` methods should be private, since they are meant to be used by the `reference_proxy` and the iterator classes only. @Klaim do you have any thought on this?

include/sparrow/layout.hpp
```diff
@@ -203,19 +208,26 @@ namespace sparrow
     }
 
     template <class T>
-    auto primitive_layout<T>::element(size_type i) -> reference
+    auto primitive_layout<T>::value(size_type i) -> inner_reference
     {
         assert(i < size());
         return i[data()];
```
JohanMabille, line 214: please, no ;) `data()[i]` is far easier to read.

JohanMabille (changes requested):
- You can consider adding `operator[]` for returning `reference_proxy` / `const_reference_proxy` objects.
- the `primitive_layout_iterator` should be updated to return `reference_proxy` / `const_reference_proxy` objects

include/sparrow/layout.hpp
Klaim: +1

include/sparrow/layout.hpp
jjerphan: So this means we cannot test the iterator anymore?

JohanMabille: You can, through the reference proxy:

```cpp
 CHECK_EQ(*iter++, lt[i].value());

File: include/sparrow/layout.hpp (hunk 1 above)
Comment (JohanMabille) [reply include/sparrow/layout.hpp:104]: these types shold be rename `inner_pointer` and `inner_const_pointer`

File: include/sparrow/layout.hpp [hunk 4]
```diff
@@ -120,13 +124,18 @@ namespace sparrow
         const_iterator cbegin() const;
         const_iterator cend() const ;
 
+        friend class reference_proxy<primitive_layout>;
+        friend class const_reference_proxy<primitive_layout>;
```
JohanMabille, line 128: these declarations shoud be in the private section.

include/sparrow/layout.hpp
```diff
@@ -108,8 +111,9 @@ namespace sparrow
         using const_iterator = primitive_layout_iterator<T, true>;
```
JohanMabille, line 111: this should be: `using value_iterator = primitive_layout_value_iterator<T, false>` and equivalent for the const iterator.

`iterator` and `const_iterator` should be the alias to the iterators over the reference proxies. These iterators should be named `primitive_layout_iterator`

include/sparrow/layout.hpp
```diff
@@ -120,13 +124,18 @@ namespace sparrow
         const_iterator cbegin() const;
         const_iterator cend() const ;
```
JohanMabille, line 125: For consistency with the itetrator names, thes methods should be renamed `value_begin` / `value_end` and `value_cbegin` / `value_cend`.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,341 @@
...
+    template <class T>
+    auto fixed_size_layout<T>::value_cbegin() const -> const_value_iterator
+    {
+        return const_value_iterator{data()};
+    }
+
+    template <class T>
+    auto fixed_size_layout<T>::value_cend() const -> const_value_iterator
+    {
+        return const_value_iterator{data() + self_type::size()};
```
JohanMabille, line 298: Nitpicking:: no need to prefix `size` with `self_type::`

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,341 @@
...
+    template <class T>
+    auto fixed_size_layout<T>::bitmap_begin() -> bitmap_iterator
+    {
+        return m_data.bitmap.begin();
+    }
+
+    template <class T>
+    auto fixed_size_layout<T>::bitmap_end() -> bitmap_iterator
+    {
+        return m_data.bitmap.begin() + self_type::size();
```
JohanMabille, line 310: Same here

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,341 @@
...
+    template <class T>
+    auto fixed_size_layout<T>::bitmap_cbegin() const -> const_bitmap_iterator
+    {
+        return m_data.bitmap.cbegin();
+    }
+
+    template <class T>
+    auto fixed_size_layout<T>::bitmap_cend() const -> const_bitmap_iterator
+    {
+        return m_data.bitmap.cbegin() + self_type::size();
```
JohanMabille, line 322: Same here.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+        using const_bitmap_range = std::ranges::subrange<const_bitmap_iterator>;
+        using const_value_range = std::ranges::subrange<const_value_iterator>;
+
+        using bitmap_range = std::ranges::subrange<bitmap_iterator>;
+        using value_range = std::ranges::subrange<value_iterator>;
+
+        // TODO: implement with `begin` and `end` once the iterator is available.
+        // using iterator = reference_proxy<self_type>::iterator;
+        // using const_iterator = const_reference_proxy<self_type>::iterator;
```
JohanMabille, line 124: That would rather be:
```cpp
using iterator = layout_iterator<self_type, false>;
using const_iterator = layout_iterator<self_type, true>;
```

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,370 @@
...
+        assert(m_data.buffers.size() > 0);
+        assert(m_data.length == m_data.buffers[0].size());
+        assert(m_data.length == m_data.bitmap.size());
+    }
+
+    template <class T>
+    auto fixed_size_layout<T>::size() const -> size_type
+    {
+        assert(m_data.buffers.size() > 0);
+        return m_data.buffers[0].size();
```
JohanMabille, line 237: I think you should read `m_data.length` instead. The reason is that `m_data.buffers[0]` may hold `uint8_t` while the data type of the array is `int32_t` for instance. Therefore, `size` would not be what you expect.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,370 @@
...
+    auto fixed_size_layout<T>::bitmap_cend() const -> const_bitmap_iterator
+    {
+        return m_data.bitmap.cbegin() + size();
+    }
+
+    template <class T>
+    auto fixed_size_layout<T>::data() -> pointer
+    {
+        assert(m_data.buffers.size() > 0);
+        return m_data.buffers[0].data();
```
JohanMabille, line 358: You should call `data<inner_value_type>` here, to be sure you cast the buffer into the right type.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    template <class T>
+    auto fixed_size_layout<T>::data() const -> const_pointer
+    {
+        assert(m_data.buffers.size() > 0);
+        return m_data.buffers[0].data();
```
JohanMabille, line 365: Same here.

test/test_layout.cpp
```diff
@@ -16,16 +16,16 @@
...
     // TODO: Test all the other base types once #15 is addressed.
-    static_assert(std::ranges::contiguous_range<primitive_layout<uint8_t>>);
-    static_assert(std::contiguous_iterator<primitive_layout_iterator<uint8_t, true>>);
-    static_assert(std::contiguous_iterator<primitive_layout_iterator<uint8_t, false>>);
+    // static_assert(std::ranges::contiguous_range<fixed_size_layout<uint8_t>>);
+    // static_assert(std::contiguous_iterator<fixed_size_layout_iterator<uint8_t, true>>);
+    // static_assert(std::contiguous_iterator<fixed_size_layout_iterator<uint8_t, false>>);
 
-    using layout_test_type = primitive_layout<uint8_t>;
+    using layout_test_type = fixed_size_layout<uint8_t>;
```
JohanMabille, line 28: I think it would be better to test the layout with `int32_t` or any primitive type different from `uint8_t`, to be sure the internal cast are correctly done everywhere.

JohanMabille (commented):
It looks like you don't take into account `m_data.offset` in the accessors and related methods.

include/sparrow/fixed_size_layout.hpp
jjerphan: Yes, those are left over from the previous design. I'll change the comment for clarity.

JohanMabille (approved):

include/sparrow/dynamic_bitset.hpp
```diff
@@ -62,7 +62,7 @@ namespace sparrow
         size_type size() const noexcept;
         size_type null_count() const noexcept;
 
-        bool test(size_type pos) const;
+        const_reference test(size_type pos) const;
```
Klaim, line 65: Could you explain why this change is necessary? To me `test` means, give me the value as of right-now, not give me access to the boolean.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+namespace sparrow
+{
+    /**
+     * An iterator for `fixed_size_layout` operating on contiguous data only.
+     *
+     * @tparam T The type of the elements in the layout's data buffer.
+     * @tparam is_const A boolean indicating whether the iterator is const.
+     *
+     * @note This class is copyable and movable.
```
Klaim, line 35: Note sure why you specified that? It's the default/usual/regular/normal/expected in C++.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,369 @@
...
+     * A contiguous layout for fixed size types.
+     *
+     * This class provides a contiguous layout for fixed size types, such as `uint8_t`, `int32_t`, etc.
+     * It iterates over the first buffer in the array_data, and uses the bitmap to skip over null.
+     * The bitmap is assumed to be present in the array_data.
+     *
+     * @tparam T The type of the elements in the layout's data buffer.
+     *           A fixed size type, such as a primitive type.
+     *
+     * @note This class is not thread-safe, exception-safe, copyable, movable, equality comparable.
```
Klaim, line 88: Again that's the default in C++ so I dont get why this is noted here?

About "exception safe" I'm not sure what you mean here, [exception safety is a scale](https://en.wikipedia.org/wiki/Exception_safety#Classification) so if it's not exception-safe "no guarantees are made" but we didnt specially designed with that in mind so to me it shouldnt even be mentionned.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -0,0 +1,369 @@
...
+        const_reference operator[](size_type i) const;
+
+        bitmap_range bitmap();
+        value_range values();
+
+        const_bitmap_range bitmap() const;
+        const_value_range values() const;
+
+    private:
+        // We only use the first buffer and the bitmap.
```
Klaim, line 140: Not sure about this comment, it might not be the best place to document this, it's kind of out of context... (thinking about it)

test/test_layout.cpp
```diff
@@ -16,68 +16,78 @@
...
-#include "sparrow/layout.hpp"
+#include "sparrow/fixed_size_layout.hpp"
 
 namespace sparrow
 {
     // TODO: Test all the other base types once #15 is addressed.
-    static_assert(std::ranges::contiguous_range<primitive_layout<uint8_t>>);
-    static_assert(std::contiguous_iterator<primitive_layout_iterator<uint8_t, true>>);
-    static_assert(std::contiguous_iterator<primitive_layout_iterator<uint8_t, false>>);
+    // static_assert(std::ranges::contiguous_range<fixed_size_layout<uint8_t>>);
```
Klaim, line 24: It cannot be a range because it doesnt expose `iterator` (which should be what `value_iterator` is) and no begin/end directly (which should be just what `value_range()` exposes). We can do that, it would make sense IMO.

include/sparrow/fixed_size_layout.hpp
jjerphan: Let me remove such notes.

include/sparrow/dynamic_bitset.hpp
jjerphan: `const_reference` is `bool` here. Changing this return type is necessary for `using`-declaration consistency which was required.

test/test_layout.cpp
jjerphan: When https://github.com/xtensor-stack/sparrow/pull/41 is merged, we should be able to adapt those assertions, yes.

include/sparrow/dynamic_bitset.hpp
Klaim: Ok so I think here `test()` should return `value_type` and, as a side note not related to this pr, `const_reference` should not be `bool` but either an actual reference to bool or a const `bitset_reference`
Anyway, here `test()` really is a capture of the current value, so `value_type` seems logical.

jjerphan: Actually yes we should remove revert this changes.

I was mislead by both the present of `test` and of `operator[]`.

Klaim: The confusion is understandable 👍🏽

Klaim (approved):

PR #32, opened 2024-03-19
Implemented variable_size_binary_layout

JohanMabille:


test/test_array_data.cpp
```diff
@@ -58,12 +60,12 @@ namespace sparrow
...
         }
 
-        inner_value_type& value(size_type index)
+        inner_reference value(size_type index)
         {
             return m_storage[index].value();
         }
 
-        const inner_value_type& value(size_type index) const
+        const inner_const_reference value(size_type index) const
```
jjerphan, line 68:
```suggestion
        inner_const_reference value(size_type index) const
```

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,93 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include "sparrow/array_data.hpp"
+#include "sparrow/iterator.hpp"
+
+namespace sparrow
+{
+    template <class T, class CR>
```
jjerphan, line 22: Can it be?
```suggestion
    template <class T, class ICR>
```

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+    /*
+     * @class variabe_size_binary_layout
+     *
+     * @brief Layout for arrays containing values consisting of a variable number of bytes.
+     *
+     * This layout is used to retrieve data in an array of values of a variable number of bytes
+     * (typically string objects). Values are stored contiguously in a data buffer, a single
+     * value is retrieved via an additional offset buffer, where each element is the beginning
+     * of the corresponinding value in the data buffer.
```
Klaim, line 106:
```suggestion
     * of the corresponding value in the data buffer.
```

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * Example:
+     *
+     * Let's consider the array ['please', 'allow', 'me', 'to', 'introduce', 'myself'].
+     * The internal buffers will be:
+     * - offset: [0, 6, 11, 13, 15, 24, 30]
+     * - data: ['p','l','e','a','s','e','a','l','l','o','w','m','e','t','o','i','n','t','r','o','d','u','c','e','m','y','s','e','l','f']
+     * 
+     * @tparam T the type of the data
```
Klaim, line 115: So in the example it is `char`? or is it `const char *` or string?

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+     * @tparam R the reference type to the data. This type is different form the reference type of the layout,
+     * which behaves like std::optional<R>.
+     * @tparam CR the const reference type to the data. This type is different from the const reference of the layout,
+     * which behaves like std::optional<CR>.
+     * @tparam OT type of the offset values. Must be std::int64_t or std::int32_t
```
Klaim, line 120: Why is this variable? An Arrow requirement?

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+     * Example:
+     *
+     * Let's consider the array ['please', 'allow', 'me', 'to', 'introduce', 'myself'].
+     * The internal buffers will be:
+     * - offset: [0, 6, 11, 13, 15, 24, 30]
+     * - data: ['p','l','e','a','s','e','a','l','l','o','w','m','e','t','o','i','n','t','r','o','d','u','c','e','m','y','s','e','l','f']
+     * 
+     * @tparam T the type of the data
+     * @tparam R the reference type to the data. This type is different form the reference type of the layout,
+     * which behaves like std::optional<R>.
```
Klaim, line 117: Consider adding a concept to constraint the reference type, as we know the kind of interface it is supposed to provide?

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+     * variable_size_binary_layout implementation *
+     **********************************************/
+
+    template <class T, class R, class CR, class OT>
+    variable_size_binary_layout<T, R, CR, OT>::variable_size_binary_layout(array_data data)
+        : m_data(std::move(data))
+    {
+        assert(m_data.buffers.size() == 2u);
+        //TODO: templatize back and front in buffer
+        //assert(m_data.buffers[0].size() == 0u || m_data.buffers[0].back() == m_data.buffers[1].size());
```
Klaim, line 238: Why is this assertion commented?

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+    template <class T, class R, class CR, class OT>
+    auto variable_size_binary_layout<T, R, CR, OT>::value(size_type i) const -> inner_const_reference
+    {
+        return inner_const_reference(data(*offset(i)), data(*offset(i+1)));
+    }
+
+    template <class T, class R, class CR, class OT>
+    auto variable_size_binary_layout<T, R, CR, OT>::offset(size_type i) const -> const_offset_iterator
+    {
+        return m_data.buffers[0].template data<OT>() + m_data.offset + i;
```
Klaim, line 292: Add an assertion that `buffers` is indeed not empty.

test/test_variable_size_binary_layout.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+        std::int64_t* offset()
+        {
+            return m_data.buffers[0].data<std::int64_t>();
+        }
+    };
+
+    TEST_SUITE("variable_size_binary_layout")
+    {
+        TEST_CASE_FIXTURE(vs_binary_fixture, "types")
```
Klaim, line 75: No need for a runtime test case when it's a bunch of compile-time checks.

I wonder if there is compile-time checks in doctest, I know there are in catch2 (I prefer doctest usually)

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+        using const_bitmap_iterator = array_data::bitmap_type::const_iterator;
+        using const_value_iterator = vs_binary_value_iterator<self_type, true>;
+
+        explicit variable_size_binary_layout(array_data data);
+
+        size_type size() const;
+        const_reference operator[](size_type i) const;
+
+        const_value_iterator value_cbegin() const;
```
Klaim, line 145: Instead of pairs of functions here, I would prefer accessors, or view accessors if it's best.

btw what is the difference between `value_cbegin()` and `begin()`?

test/test_variable_size_binary_layout.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+        array_data m_data;
+        // TODO: replace R = std::string_view with specific reference proxy
+        using layout_type = variable_size_binary_layout<std::string, std::string_view, std::string_view>;
+    
+    private:
+
+        std::int64_t* offset()
+        {
+            return m_data.buffers[0].data<std::int64_t>();
```
Klaim, line 69: add an assert checking buffers is not empty (I understand it's only tests but it needs to be resistant to change)

jjerphan (approved):

include/sparrow/variable_size_binary_layout.hpp
JohanMabille: Indeed :smiling_face_with_tear:

include/sparrow/variable_size_binary_layout.hpp
JohanMabille: We don't really know, except that an object T can be assign to an object R, the API really depends on T and can be anything.

include/sparrow/variable_size_binary_layout.hpp
JohanMabille: `value_cbegin` iterates over the data buffer, while `begin` iterates over the reference_proxy. This allows some optimizations or element wise operations when you know you don't have any missing value (because null_count is 0), or when you don't care. Notice that we only provide const iterators on the values and the bitmap to prevent inconsistency (setting a value to something while not setting a corresponding bit to true for instance).

Not sure we want accessor instead, this is really meant to speed up iterations. For value / has_value access, ref proxies are enough.

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+                is_const,
+                typename C::inner_const_reference,
+                typename C::inner_reference
+              >
+        {
+        };
+
+        template <class C, bool is_const>
+        using get_inner_reference_t = typename get_inner_reference<C, is_const>::type;
+    }
```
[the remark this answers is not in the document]
JohanMabille: I don't know yet, le'ts move it later if we need it.

include/sparrow/variable_size_binary_layout.hpp
JohanMabille: the value type is `string`. They are all stored as a contiguous buffer of char. I've added some precisions in the documentation.

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,307 @@
...
+        bool less_than(const self_type& rhs) const;
+
+        offset_iterator m_offset_it;
+        const data_iterator m_data_begin;
+
+        friend class iterator_access;
+    };
+
+    /*
+     * @class variabe_size_binary_layout
```
jjerphan, line 99:
```suggestion
     * @class variable_size_binary_layout
```

test/test_variable_size_binary_layout.cpp
Klaim: Note that if any static assert needs to use that class, then the class must be defined before that use, so probably it's best to put them next but after that class.

include/sparrow/variable_size_binary_layout.hpp
Klaim: I meant something like:
```
auto values() const; // returns a range-view using the `const_value_iterator`
```

just grouping both iterator into a view if you prefer.

include/sparrow/variable_size_binary_layout.hpp
Klaim: So it doesnt need to provide a way to know if there is a value, right?

include/sparrow/variable_size_binary_layout.hpp
Klaim: I would like a clarification in that specific line, that here the value type is the one stored in the data buffer, and not the byte representation of it.

include/sparrow/variable_size_binary_layout.hpp
Klaim: Note that if `values()` returns a range of the actual buffer, therefore with element being `char`, then I would prefer it to have another name, like `raw_data()` or something in that direction.

JohanMabille: `values` returns a range on string_view, not on char.

test/test_variable_size_binary_layout.cpp
JohanMabille: The static asserts are on the layout_type defined in the fixture class, so I guess it's fine to keep the static assert in the fixture class?

jjerphan (commented):
Yet another comment (I'll stop there).

test/test_variable_size_binary_layout.cpp
Alex-PLACET, line 1: It seems that we do not test a lot of methods because they are private.
```
reference dereference() const;
void increment();
void decrement();
void advance(difference_type n);
difference_type distance_to(const self_type& rhs) const;
bool equal(const self_type& rhs) const;
bool less_than(const self_type& rhs) const;
```
same for variable_size_binary_layout.
Do we really want to made these methods private ?

jjerphan (commented):
Mon ultime bafouille.

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,338 @@
...
+     * buffer.
+     *
+     * Example:
+     *
+     * Let's consider the array of string ['please', 'allow', 'me', 'to', 'introduce', 'myself'].
+     * The internal buffers will be:
+     * - offset: [0, 6, 11, 13, 15, 24, 30]
+     * - data: ['p','l','e','a','s','e','a','l','l','o','w','m','e','t','o','i','n','t','r','o','d','u','c','e','m','y','s','e','l','f']
+     * 
+     * @tparam T the type of the data stored in the data buffer, not its byt representation.
```
jjerphan, line 120:
```suggestion
     * @tparam T the type of the data stored in the data buffer, not its byte representation.
```

test/test_variable_size_binary_layout.cpp
JohanMabille: Actually these methods are indirectly tested by the public API of the iterator class hierarchy (CRTP). We do not want to expose them because they are really pure implementation here.

The same for the private methods of `variable_size_binary_laout`.

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -0,0 +1,338 @@
...
+
+    private:
+
+        const_value_iterator value_cbegin() const;
+        const_value_iterator value_cend() const;
+
+        const_bitmap_iterator bitmap_cbegin() const;
+        const_bitmap_iterator bitmap_cend() const;
+
+        bool has_value(size_type i) const;
```
jjerphan, line 175:
```suggestion
        bitmap_const_reference has_value(size_type i) const;
```

PR #37, opened 2024-03-21
Initialize reference proxues with references to elements instead of l…

JohanMabille:
…ayout and index

### DIFF
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -116,9 +116,11 @@ namespace sparrow
         using base_type = reference_proxy_base<self_type>;
         using layout_type = L;
         using value_type = typename L::inner_value_type;
+        using reference = typename L::inner_const_reference;
+        using bitmap_reference = typename L::bitmap_const_reference;
         using size_type = typename L::size_type;
 
-        const_reference_proxy(const layout_type& l, size_type index);
+        const_reference_proxy(reference val_ref, bitmap_reference bit_ref);
         ~const_reference_proxy() = default;
 
         const_reference_proxy(const self_type&) = default;
@@ -131,8 +133,8 @@ namespace sparrow
 
     private:
 
-        const layout_type* p_layout;
-        size_type m_index;
+        reference m_val_ref;
+        bitmap_reference m_bit_ref;
     };
 
     /**
@@ -147,9 +149,11 @@ namespace sparrow
         using base_type = reference_proxy_base<self_type>;
         using layout_type = L;
         using value_type = typename L::inner_value_type;
+        using reference = typename L::inner_reference;
+        using bitmap_reference = typename L::bitmap_reference;
         using size_type = typename L::size_type;
 
-        reference_proxy(layout_type& l, size_type index);
+        reference_proxy(reference val_ref, bitmap_reference bit_ref);
         ~reference_proxy() = default;
 
         reference_proxy(const self_type&) = default;
@@ -186,8 +190,8 @@ namespace sparrow
         template <class U>
         void update_value(U&& u);
 
-        layout_type* p_layout;
-        size_type m_index;
+        reference m_val_ref;
+        bitmap_reference m_bit_ref;
     };
 
     template <class L>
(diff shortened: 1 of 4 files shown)


include/sparrow/array_data.hpp
```diff
@@ -412,7 +417,9 @@ namespace sparrow
     template <class U>
     void reference_proxy<L>::update_value(U&& u)
     {
-        p_layout->update(m_index, std::forward<U>(u));
+        m_bit_ref = true;
```
Klaim, line 420: If the value is null, then `m_bit_ref = false` isnt it?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -309,6 +314,19 @@ namespace sparrow
         return m_null_count;
     }
 
+    template <random_access_range B>
+    auto dynamic_bitset_base<B>::operator[](size_type pos) -> reference
+    {
+        return reference(*this, m_buffer.data()[block_index(pos)], bit_mask(pos));
```
Klaim, line 320: Please add assertions about m_buffer's size vs pos :)

test/test_array_data.cpp
```diff
@@ -25,62 +25,60 @@ namespace sparrow
...
         using reference = reference_proxy<mock_layout>;
         using const_reference = const_reference_proxy<mock_layout>;
         using size_type = std::size_t;
-        using storage_type = std::vector<value_type>;
 
         mock_layout() = default;
         mock_layout(std::initializer_list<value_type> l)
-            : m_storage(l)
+            : m_bitmap(l.size())
+            , m_data(l.size())
```
Klaim, line 40: So here we are default-constructing, not sure it's the best way to do this?

test/test_array_data.cpp
```diff
@@ -25,62 +25,60 @@ namespace sparrow
...
+                {
+                    ++value_iter;
+                }
+            }
         }
 
         reference operator[](size_type pos)
         {
-            return reference(*this, pos);
+            return reference(m_data[pos], m_bitmap[pos]);
```
Klaim, line 60: Please add assertions about m_buffer's size vs pos :)

test/test_array_data.cpp
```diff
@@ -25,62 +25,60 @@ namespace sparrow
...
-        {
-            return m_storage;
-        }
-
-    private:
-
-        bool has_value(size_type index) const
-        {
-            return m_storage[index].has_value();
+            return const_reference(m_data[pos], m_bitmap[pos]);
```
Klaim, line 65: Please add assertions about m_buffer's size vs pos :)

test/test_array_data.cpp
```diff
@@ -25,62 +25,60 @@ namespace sparrow
...
         }
 
-        inner_value_type& value(size_type index)
+        bool has_value(size_type pos) const
         {
-            return m_storage[index].value();
+            return m_bitmap[pos];
```
Klaim, line 70: Please add assertions about m_buffer's size vs pos :)

test/test_array_data.cpp
```diff
@@ -25,62 +25,60 @@ namespace sparrow
...
         }
 
-        const inner_value_type& value(size_type index) const
+        inner_const_reference value(size_type pos) const
         {
-            return m_storage[index].value();
+            return m_data[pos];
```
Klaim, line 75: Please add assertions about m_buffer's size vs pos :)

include/sparrow/array_data.hpp
JohanMabille: This method updates the inner value, not the value of the array (for this one, there is `update`)

test/test_array_data.cpp
JohanMabille: I cannot think of a better alternative, but I'm opened to any suggestion ;)

jjerphan (approved):

test/test_array_data.cpp
Klaim: welp, it's tests so I guess it's fine

PR #47, opened 2024-03-27
Add clang format

### LINKED ISSUE #42: Add clang format rules
The point is to share common coding rules.

Alex-PLACET:
Fix #42
There is no warning in the checks as we only check modified files.
You can test the rules with https://clang-format-configurator.site/
There are 2 cmake targets: a dry run and an auto fix which run over all the files.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,90 @@
+BasedOnStyle: Mozilla
+
+AccessModifierOffset: '-4'
+AlignAfterOpenBracket: BlockIndent
+AlignEscapedNewlines: Left
+AllowAllArgumentsOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+# Forbid one line lambdas because clang-format makes a weird split when
+# single instructions lambdas are too long.
+AllowShortLambdasOnASingleLine: Empty
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBinaryOperators: NonAssignment
+BreakBeforeBraces: Allman
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: BeforeComma
+BreakInheritanceList: AfterComma
+BreakStringLiterals: false
+ColumnLimit: '110'
+ConstructorInitializerIndentWidth: '4'
+ContinuationIndentWidth: '4'
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat: false
+EmptyLineAfterAccessModifier: Always
+EmptyLineBeforeAccessModifier: Always
+ExperimentalAutoDetectBinPacking: true
+IncludeBlocks: Regroup
+IncludeCategories:
+- Regex: <[^.]+>
+  Priority: 1
+- Regex: <sparrow/.+>
+  Priority: 3
+- Regex: <.+>
+  Priority: 2
+- Regex: '"sparrow/.+"'
+  Priority: 4
+- Regex: '".+"'
+  Priority: 5
+IndentCaseLabels: true
+IndentWidth: '4'
+IndentWrappedFunctionNames: false
+InsertBraces: true
+InsertTrailingCommas: Wrapped
+KeepEmptyLinesAtTheStartOfBlocks: false
+LambdaBodyIndentation: Signature
+Language: Cpp
+MaxEmptyLinesToKeep: '2'
+NamespaceIndentation: All
+ObjCBlockIndentWidth: '4'
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: false
+PackConstructorInitializers: Never
+PenaltyBreakAssignment: 100000
+PenaltyBreakBeforeFirstCallParameter: 0
+PenaltyBreakComment: 10
+PenaltyBreakOpenParenthesis: 0
+PenaltyBreakTemplateDeclaration: 0
+PenaltyExcessCharacter: 10
+PenaltyIndentedWhitespace: 0
+PenaltyReturnTypeOnItsOwnLine: 10
+PointerAlignment: Left
+QualifierAlignment: Custom  # Experimental
+QualifierOrder: [inline, static, constexpr, const, volatile, type]
+ReflowComments: true
+SeparateDefinitionBlocks: Always
+SortIncludes: CaseInsensitive
+SortUsingDeclarations: true
+SpaceAfterCStyleCast: true
+SpaceAfterTemplateKeyword: true
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: '2'
+SpacesInAngles: false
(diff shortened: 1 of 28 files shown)


cmake/clang-format.cmake
```diff
@@ -0,0 +1,83 @@
...
+        message(STATUS "🪟 On Windows, you can install clang-format with:")
+        message(STATUS "\t> winget llvm")
+    endif()
+endfunction()
+
+find_program(CLANG_FORMAT clang-format
+             VALIDATOR check_clang-format_version)
+
+if(NOT CLANG_FORMAT)
+    message(WARNING "❗clang-format not found")
```
jjerphan, line 47: Nitpick.
```suggestion
    message(WARNING "❗ clang-format not found")
```

cmake/clang-format.cmake
```diff
@@ -0,0 +1,83 @@
...
+
+    print_clang_format_install_instructions()
+else()
+    get_clang_format_version(${CLANG_FORMAT})
+    message(STATUS "✅clang-format found at ${CLANG_FORMAT} | version: ${CLANG_FORMAT_MAJOR_VERSION}.${CLANG_FORMAT_MINOR_VERSION}.${CLANG_FORMAT_PATCH_VERSION}")
```
jjerphan, line 52: Nitpick.

```suggestion
    message(STATUS "✅ clang-format (version: ${CLANG_FORMAT_MAJOR_VERSION}.${CLANG_FORMAT_MINOR_VERSION}.${CLANG_FORMAT_PATCH_VERSION}") found at ${CLANG_FORMAT})
```

cmake/clang-format.cmake
```diff
@@ -0,0 +1,83 @@
...
+        message(DEBUG "clang-format found at ${item} | version: ${CLANG_FORMAT_MAJOR_VERSION}.${CLANG_FORMAT_MINOR_VERSION}.${CLANG_FORMAT_PATCH_VERSION}")
+        message(DEBUG "but version is lower than ${CLANG-FORMAT_MINIMUM_MAJOR_VERSION}")
+        set(${validator_result_var} FALSE PARENT_SCOPE)
+    else()
+        set(${validator_result_var} TRUE PARENT_SCOPE)
+    endif()
+endfunction()
+
+function(print_clang_format_install_instructions)
+    message(STATUS "🛠️ Please install clang-format to enable code formatting")
```
jjerphan, line 28: Need not be addressed: we could mention that [`clang-format` is available for most platform on conda-forge](https://prefix.dev/channels/conda-forge/packages/clang-format).

cmake/clang-format.cmake
```diff
@@ -1,36 +1,83 @@
...
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        COMMENT "Running clang-format on all files"
+    )
+
+    add_custom_target(
+        clang-format_dry_run
+        COMMAND ${CLANG_FORMAT} --dry-run -style=file ${ALL_FILES_TO_FORMAT}
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        COMMENT "Running dry clang-format on all files"
+    )
```
jjerphan, line 82: What is the purposes of running a dry run after an actual run?

Is it to check that some files cannot be formatted?

cmake/clang-format.cmake
```diff
@@ -1,36 +1,83 @@
...
+        file(GLOB_RECURSE FILES_TO_FORMAT ${CMAKE_SOURCE_DIR}/${PATTERN})
+        list(APPEND ALL_FILES_TO_FORMAT ${FILES_TO_FORMAT})
+    endforeach()
+
+    add_custom_target(
+        clang-format
+        COMMAND ${CLANG_FORMAT} -i -style=file ${ALL_FILES_TO_FORMAT}
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        COMMENT "Running clang-format on all files"
+    )
```
jjerphan, line 75: For now this target is not run.

I guess we can have another PR for a first formatting of the codebase and [ignore its merge commit for the blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).

What do you think?

jjerphan (approved):
LGTM. Thank you, @Alex-PLACET.

CMakeLists.txt
```diff
@@ -43,6 +43,15 @@ message(STATUS "Building sparrow v${${PROJECT_NAME}_VERSION}")
 
 OPTION(BUILD_TESTS "sparrow test suite" OFF)
 
+# Linter options
+# =============
+
+OPTION(ACTIVATE_LINTER "Create targets to run clang-format" OFF)
```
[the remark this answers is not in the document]
Alex-PLACET: IMO it should not be ON by default because you don't want to have as requirement to have clang-format if you just want to compile the project with GCC or MSVC (as in the CI), or as a dependency user (VCPKG, Conan)

cmake/clang-format.cmake
Alex-PLACET: A dry run will just report the place where the code is not well formated.
As de developer, the workflow is:
Run the dry run and possibly add some no-lint in the code to avoid auto format.
Run clang format auto fix

cmake/clang-format.cmake
Alex-PLACET: Yes I did not ran the formatting and I prefer to discuss the rules before to run it.

.github/workflows/analysis.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Install Ninja
+      run: sudo apt install ninja-build -y
+
+    - name: Set conda environment
+      uses: mamba-org/setup-micromamba@main
+      with:
+        environment-name: myenv
+        environment-file: environment-dev.yml
+        init-shell: bash
+        cache-downloads: true
```
jjerphan, line 30: Thinking again, I think we should only install `ninja` from conda-forge to use a single tool chain from the same prefix to prevent problem from occuring.

Since `ninja` is already part of the conda environment specification:

https://github.com/xtensor-stack/sparrow/blob/e423d1ffc35dd9cf3a277503deb6a253c59fb853/environment-dev.yml#L7

why is its installation from apt required, here?

cmake/clang-format.cmake
jjerphan: Actually you can forget about my comment since it is irrelevant: I thought those targets were run in this order as well which is not the case.

.clang-format
```diff
@@ -0,0 +1,90 @@
...
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBinaryOperators: NonAssignment
+BreakBeforeBraces: Allman
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: BeforeComma
+BreakInheritanceList: AfterComma
+BreakStringLiterals: false
+ColumnLimit: '110'
```
jjerphan, line 27: I do not have strong view for the formatting style but this perhaps constraining rule which IMO makes reading and rewriting code on GitHub and laptops easier:

```suggestion
ColumnLimit: '80'
```

JohanMabille (approved):

CMakeLists.txt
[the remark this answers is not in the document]
Klaim: also if a clang-format exists in a project's source code, msvc will take it into account automatically, vscode too assuming the right extension.

Klaim:
This PR also needs to apply the formatting rules so that it is kept in sync with the code, and to also help with seeing if we need to consider some code to not format.

.clang-format
Klaim: I dont like `80` personally because I always find it too small even on laptops, but I'm not against it if it helps Julien (just stating for the record).

jjerphan: I do not want to block this, I can adapt.

Alex-PLACET: We can update this rule later if we prefer

PR #51, opened 2024-03-27
Layouts now store a pointer to array_data instead of a value

### LINKED ISSUE #50: Layout should not own array_data
Because of `array_data` with children and recursive layouts, the layout classes should not own `array_data`, but have a reference on it. The enclosing `typed_array` should be responsible for that.

JohanMabille:
Fixes #50

### DIFF
--- a/include/sparrow/fixed_size_layout.hpp
+++ b/include/sparrow/fixed_size_layout.hpp
@@ -121,7 +121,7 @@ namespace sparrow
         // using iterator = layout_iterator<self_type, false>;
         // using const_iterator = layout_iterator<self_type, true>;
 
-        explicit fixed_size_layout(array_data p);
+        explicit fixed_size_layout(array_data& data);
 
         size_type size() const;
 
@@ -135,7 +135,8 @@ namespace sparrow
         const_value_range values() const;
 
     private:
-        array_data m_data;
+
+        array_data* p_data;
 
         pointer data();
         const_pointer data() const;
@@ -218,33 +219,33 @@ namespace sparrow
      ***********************************/
 
     template <class T>
-    fixed_size_layout<T>::fixed_size_layout(array_data ad)
-        : m_data(std::move(ad))
+    fixed_size_layout<T>::fixed_size_layout(array_data& data)
+        : p_data(&data)
     {
         // We only require the presence of the bitmap and the first buffer.
-        assert(m_data.buffers.size() > 0);
-        assert(m_data.length == m_data.bitmap.size());
+        assert(p_data->buffers.size() > 0);
+        assert(p_data->length == p_data->bitmap.size());
     }
 
     template <class T>
     auto fixed_size_layout<T>::size() const -> size_type
     {
-        assert(m_data.offset <= m_data.length);
-        return static_cast<size_type>(m_data.length - m_data.offset);
+        assert(p_data->offset <= p_data->length);
+        return static_cast<size_type>(p_data->length - p_data->offset);
     }
 
     template <class T>
     auto fixed_size_layout<T>::value(size_type i) -> inner_reference
     {
         assert(i < size());
-        return data()[i + m_data.offset];
+        return data()[i + p_data->offset];
     }
 
     template <class T>
     auto fixed_size_layout<T>::value(size_type i) const -> inner_const_reference
     {
         assert(i < size());
-        return data()[i + m_data.offset];
+        return data()[i + p_data->offset];
     }
 
     template <class T>
@@ -289,20 +290,20 @@ namespace sparrow
     auto fixed_size_layout<T>::has_value(size_type i) -> bitmap_reference
     {
         assert(i < size());
-        return m_data.bitmap[i + m_data.offset];
+        return p_data->bitmap[i + p_data->offset];
     }
 
     template <class T>
     auto fixed_size_layout<T>::has_value(size_type i) const -> bitmap_const_reference
     {
         assert(i < size());
-        return m_data.bitmap[i + m_data.offset];
+        return static_cast<const array_data*>(p_data)->bitmap[i + p_data->offset];
     }
 
     template <class T>
     auto fixed_size_layout<T>::value_begin() -> value_iterator
     {
-        return value_iterator{data() + m_data.offset};
+        return value_iterator{data() + p_data->offset};
     }
 
     template <class T>
@@ -314,7 +315,7 @@ namespace sparrow
     template <class T>
     auto fixed_size_layout<T>::value_cbegin() const -> const_value_iterator
     {
-        return const_value_iterator{data() + m_data.offset};
+        return const_value_iterator{data() + p_data->offset};
     }
 
     template <class T>
@@ -326,7 +327,7 @@ namespace sparrow
     template <class T>
     auto fixed_size_layout<T>::bitmap_begin() -> bitmap_iterator
     {
-        return m_data.bitmap.begin() + m_data.offset;
+        return p_data->bitmap.begin() + p_data->offset;
     }
 
     template <class T>
@@ -338,7 +339,7 @@ namespace sparrow
     template <class T>
     auto fixed_size_layout<T>::bitmap_cbegin() const -> const_bitmap_iterator
     {
-        return m_data.bitmap.cbegin() + m_data.offset;
+        return p_data->bitmap.cbegin() + p_data->offset;
     }
 
     template <class T>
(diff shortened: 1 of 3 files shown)


include/sparrow/fixed_size_layout.hpp
```diff
@@ -289,20 +290,20 @@ namespace sparrow
...
-        return m_data.bitmap[i + m_data.offset];
+        return p_data->bitmap[i + p_data->offset];
     }
 
     template <class T>
     auto fixed_size_layout<T>::has_value(size_type i) const -> bitmap_const_reference
     {
         assert(i < size());
-        return m_data.bitmap[i + m_data.offset];
+        return static_cast<const array_data*>(p_data)->bitmap[i + p_data->offset];
```
Klaim, line 300: Is this a `const_cast`? not sure why there is a need for a cast here

include/sparrow/fixed_size_layout.hpp
```diff
@@ -121,7 +121,7 @@ namespace sparrow
         // using iterator = layout_iterator<self_type, false>;
         // using const_iterator = layout_iterator<self_type, true>;
 
-        explicit fixed_size_layout(array_data p);
+        explicit fixed_size_layout(array_data& data);
```
Klaim, line 124: In general, can layouts be moved or copied? My assumption is yes but if so we need to add ways for these operations to be safe, in particular in the context of moving `array_data` objects around (if that would happen). Because now that it's a pointer being stored, moved-from layouts might lead to UB?

include/sparrow/fixed_size_layout.hpp
```diff
@@ -218,33 +219,33 @@ namespace sparrow
...
     template <class T>
-    fixed_size_layout<T>::fixed_size_layout(array_data ad)
-        : m_data(std::move(ad))
+    fixed_size_layout<T>::fixed_size_layout(array_data& data)
+        : p_data(&data)
     {
         // We only require the presence of the bitmap and the first buffer.
-        assert(m_data.buffers.size() > 0);
-        assert(m_data.length == m_data.bitmap.size());
+        assert(p_data->buffers.size() > 0);
```
Klaim, line 226: We might need a free function checking if an `array_data` seems well-formed, and use that in asserts too.

include/sparrow/fixed_size_layout.hpp
JohanMabille: I don't think so, usually `const_cast` is used to convert a const type to a non const one.
This is required otherwise `p_data` being `array_data*`, `p_array_data->bitmap[...]` return a `bitmap_reference` instead of a `bitmap_const_reference`.

include/sparrow/fixed_size_layout.hpp
JohanMabille: Ah good catch, the overall structure would be something like:
```cpp
tempate < .... >
class typed_array
{

private:
    array_data m_data;
    layout_type m_layout;
];
```
With m_layout referencing `m_data`, and everything being movable / copiable ...

JohanMabille: One solution would be to move `m_data` into a `std::unique_ptr` to guarantee in-memory stability.

JohanMabille: A third option is to delete copy and move semantics of layouts and reinstantiate them each time we copy / move a `typed_array`. This is actually my favorite option.

include/sparrow/fixed_size_layout.hpp
```diff
@@ -135,7 +140,8 @@ namespace sparrow
         const_value_range values() const;
 
     private:
-        array_data m_data;
+
+        array_data* p_data;
```
Alex-PLACET, line 144: Why storing the pointer and not the reference ?

JohanMabille: It's more flexible:
- the layout can become easily default constructible if we need it
- we can assign to a pointer, not to a reference
- this storage is private, so it's fine to use a pointer instead of a reference..

include/sparrow/fixed_size_layout.hpp
JohanMabille: Whether an `array_data` is well-formed is actually specific to the layout (even if some layouts share some invariants).

include/sparrow/fixed_size_layout.hpp
Alex-PLACET: What about using std::reference_wrapper ?
My point is that a reference explicits the ownership. It's less clear with a pointer.

JohanMabille: That's a good point. Let me update the PR.

Alex-PLACET (approved):

include/sparrow/fixed_size_layout.hpp
Klaim: > I don't think so, usually const_cast is used to convert a const type to a non const one.

Yes but it's also sending the right message about what the cast is about, whatever the direction.

I see it's the pointed type that you need to make const ok, then I would prefer something clearer like:
```c++
const auto& readonly_data = *p_data;
return readonly_data.bitmap[i+readonly_data.offset];
```
That way the intention is clear. `static_cast` here sends the wrong message.

include/sparrow/fixed_size_layout.hpp
Klaim: So that leads to questions like:
- if we put a `unique_ptr` in `typed_array` , is the cost of allocation a problem? (assuming both members would be contained in the allocated object, like a pimpl)
- is it ok to implement `typed_array`'s copy operations manually? what happens to `m_layout` if it's not copyable but we need to copy the `typed_array`?

I feel like keeping layouts copy and move enabled might be the simplest solution, not sure...

include/sparrow/fixed_size_layout.hpp
Klaim: In that case it might be necessary to make a concept for layouts that requests, among other thing, that these layout types provide each a way to check if the provided `array_data` matches the expected structure.
Then, because of the nature of the construction of the layout, the check must be done whatever the build context and then we must throw an exception (or abort, that depends on what we decide to support as error handling) or keep assertions and *document*  that providing an `array_data` to constructor layout that doesnt match the expected structure is undefined (and maybe provide a way to activate the check).
Lots of tradeoff choices here.

include/sparrow/fixed_size_layout.hpp
JohanMabille: Agreed. Notice that with the `std::reference_wrapper` instead of the pointer, we don't have this issue anymore.

include/sparrow/fixed_size_layout.hpp
JohanMabille: The copy / move constructor of `typed_array` would be something like:
```cpp
typed_array(const typed_array& rhs)
    : m_data(rhs.m_data)
    , m_layout(m_data)
{
}

typed_array(typed_array&& rhs)
    : m_data(std::move(rhs.m_data))
    , m_layout(m_data)
{
}
```

The issue is for assignment operators (copy and move). We probably need an additional "reset_array_data" method to be able to implement these operators.

The issue with keeping the copy and move semantic is that it will be sliently wrong the layout ogf a copied array will still refer to the array_data of the initial array.

> if we put a unique_ptr in typed_array , is the cost of allocation a problem? (assuming both members would be contained in the allocated object, like a pimpl)

I don't really know; I would say that as much as we can avoid dynamic allocation, we should do it. Especially if the copy / move semantic of `typed_array` is "easy enough" to implement without the unique_ptr (even if I admit that "easy enough" is totally suggestive)

Klaim: > I would say that as much as we can avoid dynamic allocation, we should do it. 

I agree.

> The issue is for assignment operators (copy and move). We probably need an additional "reset_array_data" method to be able to implement these operators.

In those cases, only the `m_data` should be moved and copied, as `m_layout` always refers to the local `m_data`. No need to change `m_layout`... Or am I missing something? Does `m_layout` have an actual state? Depends on the layout type?

JohanMabille: The layout classes should be stateless. 

And yes, only the `array_data` member should be copied / moved. But then the layout member must be "built" again to refer the "new" array_data.

include/sparrow/fixed_size_layout.hpp
JohanMabille: I think we should do this in a dedicated PR. I'll open an issue to track it.

Klaim (approved):
LGTM, dont forget to note the things to do in future prs

PR #53, opened 2024-03-29
Implemented pointer_iterator

JohanMabille:
The current `buffer_iterator` and `fixed_size_layout_value_iterator` are very similar, they adapt a pointer to make it an iterator. More classes to come will have to do the same. This PR captures the common implementation in the `pointer_iterator` class, to reduce code duplication.

### DIFF
--- a/include/sparrow/iterator.hpp
+++ b/include/sparrow/iterator.hpp
@@ -401,4 +401,88 @@ namespace sparrow
         : public iterator_root_base<Derived, Element, IteratorConcept, Reference, Difference>
     {
     };
+
+    /*
+     * @class pointer_iterator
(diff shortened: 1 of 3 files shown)


Alex-PLACET (approved):

include/sparrow/iterator.hpp
```diff
@@ -401,4 +401,88 @@ namespace sparrow
         : public iterator_root_base<Derived, Element, IteratorConcept, Reference, Difference>
     {
     };
+
+    /*
+     * @class pointer_iterator
```
Klaim, line 406: you can remove the `@class ...` and rely on autobrief 🥲

Klaim (approved):

PR #55, opened 2024-04-01
Implemented type erasure for allocators

### LINKED ISSUE #54: Implement allocator type erasure
This will be used when refactoring the buffer class. It will allow the user to pass any allocator to the buffer while not propagating an allocator template parameter from the buffer class.

JohanMabille:
Fixes #54

### DIFF
--- a/include/sparrow/allocator.hpp
+++ b/include/sparrow/allocator.hpp
@@ -57,16 +57,8 @@ namespace sparrow
 
         using value_type = T;
 
-        any_allocator()
-            : m_storage(make_storage(DA()))
-        {
-        }
-
-        any_allocator(const any_allocator& rhs)
-            : m_storage(copy_storage(rhs.m_storage))
-        {
-        }
-
+        any_allocator();
+        any_allocator(const any_allocator& rhs);
         any_allocator(any_allocator&&) = default;
 
         any_allocator& operator=(const any_allocator& rhs) = delete;
@@ -80,66 +72,12 @@ namespace sparrow
         {
         }
 
-        [[nodiscard]] T* allocate(std::size_t n)
-        {
-            return std::visit([n](auto&& arg)
-            {
-                using A = std::decay_t<decltype(arg)>;
-                if constexpr (is_value_stored<A>)
-                    return arg.allocate(n);
-                else
-                    return arg->allocate(n);
-            }, m_storage);
-        }
+        [[nodiscard]] T* allocate(std::size_t n);
+        void deallocate(T* p, std::size_t n);
+        
+        any_allocator select_on_container_copy_construction() const;
 
-        void deallocate(T* p, std::size_t n)
-        {
-            return std::visit([p, n](auto&& arg)
-            {
-                using A = std::decay_t<decltype(arg)>;
-                if constexpr (is_value_stored<A>)
-                    arg.deallocate(p, n);
-                else
-                    arg->deallocate(p, n);
-            }, m_storage);
-        }
-
-        any_allocator select_on_container_copy_construction() const
-        {
-            return any_allocator(*this);
-        }
-
-        bool equal(const any_allocator& rhs) const
-        {
-            // YOLO!!
-            return std::visit([&rhs](auto&& arg)
-            {
-                using A = std::decay_t<decltype(arg)>;
-                if constexpr (is_value_stored<A>)
-                {
-                    return std::visit([&arg](auto&& arg2)
-                    {
-                        using A2 = std::decay_t<decltype(arg2)>;
-                        if constexpr (is_value_stored<A2> && std::same_as<A, A2>)
-                            return arg == arg2;
-                        else
-                            return false;
-                    }, rhs.m_storage);
-                }
-                else
-                {
-                    return std::visit([&arg](auto&& arg2)
-                    {
-                        using A2 = std::decay_t<decltype(arg2)>;
-                        if constexpr (is_value_stored<A2>)
-                            return false;
-                        else
-                            return arg->equal(*arg2);
-                    }, rhs.m_storage);
-                }
-
-            }, m_storage);
-        }
+        bool equal(const any_allocator& rhs) const; 
 
     private:
 
@@ -218,6 +156,87 @@ namespace sparrow
 
         storage_type m_storage;
     };
+
+    /********************************
+     * any_allocator implementation *
+     ********************************/
+
+    template <class T, class DA>
+    any_allocator<T, DA>::any_allocator()
+        : m_storage(make_storage(DA()))
+    {
+    }
+
+    template <class T, class DA>
+    any_allocator<T, DA>::any_allocator(const any_allocator& rhs)
+        : m_storage(copy_storage(rhs.m_storage))
+    {
+    }
+
+    template <class T, class DA>
+    [[nodiscard]] T* any_allocator<T, DA>::allocate(std::size_t n)
+    {
+        return std::visit([n](auto&& arg)
+        {
+            using A = std::decay_t<decltype(arg)>;
+            if constexpr (is_value_stored<A>)
+                return arg.allocate(n);
+            else
+                return arg->allocate(n);
+        }, m_storage);
+    }
+
+    template <class T, class DA>
+    void any_allocator<T, DA>::deallocate(T* p, std::size_t n)
+    {
+        return std::visit([p, n](auto&& arg)
+        {
+            using A = std::decay_t<decltype(arg)>;
+            if constexpr (is_value_stored<A>)
+                arg.deallocate(p, n);
+            else
+                arg->deallocate(p, n);
+        }, m_storage);
+    }
+
+    template <class T, class DA>
+    any_allocator<T, DA> any_allocator<T, DA>::select_on_container_copy_construction() const
+    {
+        return any_allocator(*this);
+    }
+
+    template <class T, class DA>
+    bool any_allocator<T, DA>::equal(const any_allocator& rhs) const
+    {
+        // YOLO!!
+        return std::visit([&rhs](auto&& arg)
+        {
+            using A = std::decay_t<decltype(arg)>;
+            if constexpr (is_value_stored<A>)
+            {
+                return std::visit([&arg](auto&& arg2)
+                {
+                    using A2 = std::decay_t<decltype(arg2)>;
+                    if constexpr (is_value_stored<A2> && std::same_as<A, A2>)
+                        return arg == arg2;
+                    else
+                        return false;
+                }, rhs.m_storage);
+            }
+            else
+            {
+                return std::visit([&arg](auto&& arg2)
+                {
+                    using A2 = std::decay_t<decltype(arg2)>;
+                    if constexpr (is_value_stored<A2>)
+                        return false;
+                    else
+                        return arg->equal(*arg2);
+                }, rhs.m_storage);
+            }
+
+        }, m_storage);
+    }
 
     template <class T, class DA>
     bool operator==(const any_allocator<T, DA>& lhs, const any_allocator<T, DA>& rhs)
--- a/include/sparrow/allocator.hpp
+++ b/include/sparrow/allocator.hpp
@@ -147,9 +147,10 @@ namespace sparrow
             return std::forward<A>(alloc);
         }
 
-        template <class... Ts>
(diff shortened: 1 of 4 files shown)


jjerphan (approved):

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,148 @@
...
+            {
+                if (auto* p = dynamic_cast<const impl<A>*>(&rhs))
+                {
+                    return p->m_alloc == m_alloc;
+                }
+                return false;
+            }
+        };
+
+        std::unique_ptr<interface> p_allocator;
```
Klaim, line 139: While we agreed to start with minimal effort for this implementation, I believe we might want to immediately support either small buffer optimization (there are various ways to do this but maybe using `std::variant`, `std::any` or `std::function` - the last 2 ones already do sbo) OR just support at least that there is no allocation (no `make_unique`) when the allocator type is actually the default (`DA`) by then just storing it as a value (maybe using a `variant<D1, unique_ptr>`?).

My main worry is the cost in "normal" case where `DA` is `std::allocator`.

test/test_allocator.cpp
```diff
@@ -0,0 +1,96 @@
...
+        value_type* buf = a.allocate(n);
+        std::uninitialized_copy(ref.cbegin(), ref.cend(), buf);
+        CHECK_EQ(*buf, ref[0]);
+        CHECK_EQ(*(buf + n - 1), ref.back());
+        a.deallocate(buf, n);
+    }
+
+#if __APPLE__
+    // /usr/lib/libc++.1.dylib is missing the symbol __ZNSt3__13pmr20get_default_resourceEv, leading
+    // to an exception at runtime.
```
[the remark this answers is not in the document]
Klaim: My understanding is that apple-clang still today do not provide pmr anyway?

test/test_allocator.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+            A alloc;
+            sparrow::any_allocator<value_type> a(alloc);
+            sparrow::any_allocator<value_type> b(a);
+            CHECK(a == b);
+
+            sparrow::any_allocator<value_type> c(alloc);
+            b = c;
+            CHECK(b == c);
+        }
```
Klaim, line 51: Side comment, in case: I dont know if it's possible to check with allocators, but usually checking copy operations implies checking that modifying one of the copy will not affect the others. Not sure it makes sense here, because an allocator could be designed to be an access point to an Arena or Pool.

JohanMabille: Indeed it does not really make sense here, as `std::allocator` is stateless, and you can not change the resource pointer of a `polymorphic_allocator` when it's built.

test/test_allocator.cpp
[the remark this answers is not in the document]
JohanMabille: The weird thing is the code compiles (and links), but the symbol is missing at `runtime`. For me, it's a bug in their implementation of the standard library.

include/sparrow/allocator.hpp
JohanMabille: Indeed, let's fix this issue before this gets in.

JohanMabille:
Notice that I had to delete assignment operators, since `polymorphic_allocator` is actually not assignable.

test/test_allocator.cpp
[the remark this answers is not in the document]
jjerphan: Or it might be that the shared library at link time is missing the symbol because of a header mismatch.

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+        {
+        }
+
+        [[nodiscard]] T* allocate(std::size_t n);
+        void deallocate(T* p, std::size_t n);
+        
+        any_allocator select_on_container_copy_construction() const;
+
+        bool equal(const any_allocator& rhs) const; 
+
```
Klaim, line 81: Consider adding
```c++
contexpr std::type_index typeid() const;
```
So that the type can be introspectable in a cheap way.

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+            }
+
+            std::unique_ptr<interface> clone() const override
+            {
+                return std::make_unique<impl<A>>(m_alloc);
+            }
+
+            bool equal(const interface& rhs) const override
+            {
+                if (auto* p = dynamic_cast<const impl<A>*>(&rhs))
```
Klaim, line 117: Invoking `dynamic_cast` is super expensive and must be avoided at all costs.
Could a solution using `std::type_index(typeid(T))` work?

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+                {
+                    return p->m_alloc == m_alloc;
+                }
+                return false;
+            }
+        };
+
+        using storage_type = std::variant
+        <
+            std::allocator<T>,
```
Klaim, line 127: hmm if we had a `unique` meta-prog algorithm , you could add `DA` to the list, and that way we also optimize for `DA`.

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+     * (standard, polymorphic) without having to expose it as a template parameter.
+     *
+     * @tparam T value_type of the allocator
+     * @tparam DA default allocator, instantiated when calling the default constructor
+     */
+    template <class T, class DA = std::allocator<T>>
+    class any_allocator
+    {
+        template <class A>
+        static constexpr bool is_value_stored =
```
Klaim, line 53: - this is essentially a concept, I think it would be cheaper as a concept as it wouldnt exist in memory at all and might avoid generating unnecessary symbols - that would mean moving this outside but this doesnt seem to be an issue to me; 
- the name is not clear as to what it means (at least I dont understand it compared to what it checks);
- why check only for standard allocators and not `DA`?

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+    any_allocator<T, DA>::any_allocator(const any_allocator& rhs)
+        : m_storage(copy_storage(rhs.m_storage))
+    {
+    }
+
+    template <class T, class DA>
+    [[nodiscard]] T* any_allocator<T, DA>::allocate(std::size_t n)
+    {
+        return std::visit([n](auto&& arg)
+        {
```
Klaim, line 180: looks like thius lambda is repeated partially, maybe ther eis a way to factorize this?

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    template <class T, class DA>
+    any_allocator<T, DA> any_allocator<T, DA>::select_on_container_copy_construction() const
+    {
+        return any_allocator(*this);
+    }
+
+    template <class T, class DA>
+    bool any_allocator<T, DA>::equal(const any_allocator& rhs) const
```
Klaim, line 209: I'm not sure we even need allocator comparison? Usually it's only useful to be able to see what's the type of the allocator, it doesnt really have an observable state (at least standard allocators dont).

Klaim: ok that's part of standard requiremetns

test/test_allocator.cpp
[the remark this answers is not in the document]
jjerphan: It seems that this problem has been reported via [this thread in the Apple Developers Forum](https://forums.developer.apple.com/forums/thread/747547).

> My understanding is that apple-clang still today do not provide pmr anyway?

Do you have a reference on that, @Klaim?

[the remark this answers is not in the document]
Klaim: Only the usual rant in various communities and my own experience, but things might have changed: cppreference says
image

include/sparrow/allocator.hpp
JohanMabille: I'm not sure we need it as a public API. What the `any_allocator` type really holds should remain private in my opinion.

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+                     typename std::remove_cvref_t<T>::value_type* p,
+                     std::size_t n)
+        {
+            typename std::remove_cvref_t<T>::value_type;
+            { alloc.allocate(n) } -> std::same_as<typename std::remove_cvref_t<T>::value_type*>;
+            { alloc.deallocate(p, n) } -> std::same_as<void>;
+        };
+
+    template <class A, class T>
+    concept can_any_allocator_sbo = allocator<A> &&
```
jjerphan, line 44: Could you document what this concept allows?

jjerphan (approved):
LGTM module a few comments for documentation.

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,247 @@
...
+            using A = std::decay_t<decltype(arg)>;
+            if constexpr (can_any_allocator_sbo<A, T>)
+                arg.deallocate(p, n);
+            else
+                arg->deallocate(p, n);
+        }, m_storage);
+    }
+
+    template <class T>
+    any_allocator<T> any_allocator<T>::select_on_container_copy_construction() const
```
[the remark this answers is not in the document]
JohanMabille: It's imposed by the [standard](https://en.cppreference.com/w/cpp/memory/allocator_traits/select_on_container_copy_construction)

include/sparrow/allocator.hpp
JohanMabille: In that case I don't see any mean to factorize the code without making it overcomplicated: you would need to pass two additional lambdas to a common function. I find it a bit of overkill to factorize out a simple if.

include/sparrow/allocator.hpp
Klaim: Wouldn't user code that need to transfer the data at some point need to know that information? not sure.

include/sparrow/allocator.hpp
Klaim: The only thing that change in these function is the syntax to access the stored object. So a general accessor which provide a common interface whatever the type would solve the issue. Something to be able to change this code to
```c++
return some_nice_name([n](auto& allocator){ return allocator.allocate(n); };
```

I disagree that it's overcomplicated (to have that accessor/mutator/visitor function) because I find the current access code repeated obfuscating the important part of these functions: we just want to call allocate and deallocate here, whatever the type.

JohanMabille: > I disagree that it's overcomplicated

Indeed, what I had in mind was factorizing the control flow in a third function that would be called by these two methods. Providing accessors sohuld actually be quite trivial

JohanMabille: > I disagree that it's overcomplicated

Indeed, it's not the "design" I had in mind when I commented. Factorieing the code with a general accessor should be easier.

include/sparrow/allocator.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+        template <class... Ts>
+        struct overloaded : Ts... { using Ts::operator()...; };
+
+        storage_type copy_storage(const storage_type& rhs) const
+        {
+            return std::visit(overloaded {
+                [](const auto& arg) -> storage_type { return { std::decay_t<decltype(arg)>(arg) }; },
+                [](const std::unique_ptr<interface>& arg) -> storage_type { return { arg->clone() }; }
+            }, rhs);
```
JohanMabille, line 158: Change done, I haven't reuse `visit_storage` in the implementation of `copy_storage` for the following reasons:
- copy_storage is const, while allocate and deallocate are not
- copy_storage operates on another storage than `m_storage`

=> This means we need a second overload of the `visit_storage` method to have the second lambda accept `const interface&` instead of `const std::unique_ptr<interface>&`. I'm not sure adding so much code for changing an argument is worth it  (i.e. I find the code understandable without it), but happy to do it if you think so.

Klaim: I feel it's fine like this

Klaim (approved):
LGTM

test/test_allocator.cpp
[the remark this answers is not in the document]
jjerphan: Using Xcode 15 should allow supporting more features, including `std::pmr::polymorphic_allocator`: https://developer.apple.com/xcode/cpp/#c++20

PR #59, opened 2024-04-04
Add value_ptr

### LINKED ISSUE #57: Implement a value_ptr class
This class should be similar to a `unique_ptr`, but copiable: it should copy the pointed object.
This can be implemented in a `memory.hpp` header, where we can transfer the allocator implementation when #55 is merged.

Alex-PLACET:
Fix #57 
Implement value_ptr class

### DIFF
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -48,7 +48,7 @@ namespace sparrow
         // Other buffers
         std::vector<buffer_type> buffers;
         std::vector<array_data> child_data;
-        sparrow::value_ptr<array_data> dictionary;
+        value_ptr<array_data> dictionary;
     };
 
     /**
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -16,13 +16,13 @@
 
 #include <cassert>
 #include <compare>
-#include <concepts>
 #include <optional>
 #include <vector>
 
 #include "sparrow/buffer.hpp"
 #include "sparrow/data_type.hpp"
 #include "sparrow/dynamic_bitset.hpp"
+#include "sparrow/memory.hpp"
 
 namespace sparrow
 {
@@ -48,7 +48,7 @@ namespace sparrow
         // Other buffers
         std::vector<buffer_type> buffers;
         std::vector<array_data> child_data;
-        std::shared_ptr<array_data> dictionary;
+        sparrow::value_ptr<array_data> dictionary;
     };
 
     /**
--- a/include/sparrow/memory.hpp
+++ b/include/sparrow/memory.hpp
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <cassert>
 #include <memory>
 
--- a/include/sparrow/memory.hpp
+++ b/include/sparrow/memory.hpp
@@ -1,0 +1,83 @@
+#include <stdexcept>
+#include <memory>
+
+namespace sparrow {
+
+template <class T>
+class value_ptr {
+public:
+    value_ptr() = default;
+
+    explicit value_ptr(T value)
+        : value_(std::make_unique<T>(std::move(value))) {}
+
+    value_ptr(const value_ptr& other)
+        : value_(other.value_ ? std::make_unique<T>(*other.value_) : std::unique_ptr<T>()){}
+
+    value_ptr(value_ptr&& other) noexcept
+        : value_(std::move(other.value_)) {
+        other.reset();
+    }
+
+    value_ptr& operator=(const value_ptr& other) {
+        if (other.has_value()) {
+            value_ = std::make_unique<T>(*other.value_);
+        }else {
+            value_.reset();
+        }
+        return *this;
+    }
+
+    value_ptr& operator=(value_ptr&& other) noexcept {
+        if (this != &other) {
+            value_ = std::move(other.value_);
+            other.reset();
+        }
+        return *this;
+    }
+
+    T& operator*() {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
+        }
+        return *value_;
+    }
+
+    const T& operator*() const {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
+        }
+        return *value_;
+    }
+
+    T* operator->() {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
+        }
+        return &*value_;
+    }
+
+    const T* operator->() const {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
+        }
+        return &*value_;
+    }
+
+    explicit operator bool() const noexcept {
+        return has_value();
+    }
+
+    bool has_value() const noexcept {
+        return bool(value_);
+    }
+
+    void reset() noexcept {
+        value_.reset();
+    }
+
+private:
+    std::unique_ptr<T> value_;
+};
+
+}
--- a/include/sparrow/memory.hpp
+++ b/include/sparrow/memory.hpp
@@ -17,84 +17,107 @@
 #include <cassert>
 #include <memory>
 
-namespace sparrow {
+namespace sparrow
+{
 
-/**
- * @brief A value_ptr is a smart pointer that behaves like a value.
- * It manages the lifetime of an object of type T which is not stored in the `value_ptr` but a pointer, similar to `unique_ptr`.
- * When copied, it copies the managed object.
- * 
- * @tparam T The type of the object managed by the `value_ptr`.
- * @todo Make it constexpr.
- */
-template <class T>
-class value_ptr {
-public:
-    value_ptr() = default;
+    /**
+     * @brief A value_ptr is a smart pointer that behaves like a value.
+     * It manages the lifetime of an object of type T which is not stored in the `value_ptr` but a pointer,
+     * similar to `unique_ptr`. When copied, it copies the managed object.
+     *
+     * @tparam T The type of the object managed by the `value_ptr`.
+     * @todo Make it constexpr.
+     */
+    template <class T>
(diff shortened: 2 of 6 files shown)


include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include <memory>
+
+namespace sparrow {
+
+template <class T>
+class value_ptr {
+public:
+    value_ptr() = default;
+
+    explicit value_ptr(T value)
```
JohanMabille, line 11: I think the constructor should take a pointer instead of a value, to avoid unecessary copy when we want `value_ptr` to capture an existing pointer. A free function `make_value_ptr` can accept a value and internally allcoate a pointer and pass it to the `value_ptr` (similar to `std::make_unique_ptr`).

JohanMabille (commented):
Formatting: the opening curly braces should go on a new line

include/sparrow/array_data.hpp
```diff
@@ -48,7 +48,7 @@ namespace sparrow
         // Other buffers
         std::vector<buffer_type> buffers;
         std::vector<array_data> child_data;
-        std::shared_ptr<array_data> dictionary;
+        sparrow::value_ptr<array_data> dictionary;
```
Klaim, line 51: We already are in the `sparrow` namespace, you can remove `sparrow::`

include/sparrow/memory.hpp
Klaim: I disagree, it should be able to take a pointer or a value. Both cases are useful.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+        : value_(std::make_unique<T>(std::move(value))) {}
+
+    value_ptr(const value_ptr& other)
+        : value_(other.value_ ? std::make_unique<T>(*other.value_) : std::unique_ptr<T>()){}
+
+    value_ptr(value_ptr&& other) noexcept
+        : value_(std::move(other.value_)) {
+        other.reset();
```
Klaim, line 19: I suspect this is not necessary because move operation are specified to use `.release()` on moved-from unique_ptr, which means `other` is already set to `nullptr` after the move.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    value_ptr& operator=(const value_ptr& other) {
+        if (other.has_value()) {
+            value_ = std::make_unique<T>(*other.value_);
```
Klaim, line 24: I think a test is missing here: if `value_` has value, you need to copy `other.value_` in it, not re-allocate anything.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+        }else {
+            value_.reset();
+        }
+        return *this;
+    }
+
+    value_ptr& operator=(value_ptr&& other) noexcept {
+        if (this != &other) {
+            value_ = std::move(other.value_);
+            other.reset();
```
Klaim, line 34: Same as previously: move assignation for unique_ptr is specified as if `other.release()` was called, so just doing the move will be enough.

Actually for both move operations, I think you can just use ` = default`?

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return *this;
+    }
+
+    T& operator*() {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
```
Klaim, line 41: That should not be a runtime check but an assertion/contract: add a documented precondition that this operator requires that `has_value() == true`.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return *value_;
+    }
+
+    const T& operator*() const {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
```
Klaim, line 48: Same contract/assert instea of runtime error issue as before.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return *value_;
+    }
+
+    T* operator->() {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
```
Klaim, line 55: Same contract/assert instea of runtime error issue as before.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return &*value_;
+    }
+
+    const T* operator->() const {
+        if (!value_) {
+            throw std::runtime_error("No value stored in value_ptr");
```
Klaim, line 62: Same contract/assert instea of runtime error issue as before.

test/test_memory.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+        *vp = 43;
+        CHECK(*vp == 43);
+    }
+
+    TEST_CASE("operator->")
+    {
+        value_ptr vp(std::vector<int>{42});
+        CHECK(vp.operator->() == &*vp);
+        CHECK(vp->size() == 1);
```
Klaim, line 81: Consider adding
```
CHECK(vp->at(0) == 42);
```
to be sure observable value is what we expect.

include/sparrow/memory.hpp
Alex-PLACET: You're riht, default is enough

include/sparrow/memory.hpp
JohanMabille: Yeah, my point was that constructing form a value could be done via a free function, to be homogeneous with the STL. But as long as both cases are supported (ivia construcrors or constructor + free function) , I'm fine.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,71 @@
+#include <cassert>
+#include <memory>
+
+namespace sparrow {
+
+template <class T>
+class value_ptr {
```
jjerphan, line 7: It is possible to document the motivation behind `value_ptr`?

test/test_memory.cpp
```diff
@@ -0,0 +1,110 @@
...
+        CHECK(*vp == 42);
+
+        *vp = 43;
+        CHECK(*vp == 43);
+    }
+
+    TEST_CASE("operator->")
+    {
+        value_ptr vp(std::vector<int>{42});
+        CHECK(vp.operator->() == &*vp);
```
jjerphan, line 80: I believe the following is slightly safer and explicit since `operator&` can be overloaded.
```suggestion
        CHECK(vp.operator->() == std::addressof(*vp));
```

jjerphan (approved):
LGTM.

I mainly have a minor suggestion.

JohanMabille:
I think `test_memory.cpp` should also have a copyright.

include/sparrow/memory.hpp
Klaim: Not sure what would be the utility of such free function, it makes sense for shared and unique ptr because they dont do somethnig trivial, but that's not the case here.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,94 @@
...
+#pragma once
+
+#include <cassert>
+#include <memory>
+
+namespace sparrow {
+
+/**
+ * @brief A value_ptr is a smart pointer that behaves like a value.
+ * It manages the lifetime of an object of type T.
```
Klaim, line 24:
```suggestion
 * It manages the lifetime of an object of type T which is not stored in the `value_ptr` but a pointer, similar to `unique_ptr`.
```

or some clarification in that direction.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,94 @@
...
+    value_ptr(const value_ptr& other)
+        : value_(other.value_ ? std::make_unique<T>(*other.value_) : std::unique_ptr<T>()){}
+
+    value_ptr(value_ptr&& other) noexcept = default;
+
+    ~value_ptr() = default;
+
+    value_ptr& operator=(const value_ptr& other) {
+        if (other.has_value()) {
+            *value_ = *other.value_;
```
Klaim, line 49:
```suggestion
            if(value_)
            {
                *value_ = *other.value_;
            }
            else
            {
                value_ = std::make_unique(*other.value_);
            }
```

The code before my suggestion is ub if `value_` is `nullptr`, you need to check for it too and only `if(value_ && other.has_value())` can you do the copy of pointed values. If `not value_` then you need to do the code you did previously with make_unique of the value of other.

include/sparrow/memory.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * @brief A value_ptr is a smart pointer that behaves like a value.
+ * It manages the lifetime of an object of type T.
+ * When copied, it copies the managed object.
+ * 
+ * @tparam T The type of the object managed by the `value_ptr`.
+ */
+template <class T>
+class value_ptr {
+public:
+    value_ptr() = default;
```
Klaim, line 32: `unique_ptr` is `constexpr` in C++23, I believe this type can also be written all `constexpr` although it might not work because of the calls to `make_unique`. Anyway, an attempt to make this type all `constexpr` would be helpful in the future.

include/sparrow/memory.hpp
Alex-PLACET: Good catch

Alex-PLACET: I added UT too

include/sparrow/memory.hpp
jjerphan: We only can support C++20 unfortunately, is that it?

SylvainCorlay: This could be (locally) set to constexpr with a macro only defined when compiling with C++23.

JohanMabille (approved):

Issue #61, opened 2024-04-04
Implement `typed_array`


jjerphan:
A layout class and the `array_data` structure are gathered into the `typed_array` template class, which provides an API very similar to that of `std::vector`:

```cpp
template <class T, class L = get_default_layout_t<T>>
class typed_array
{
public:
    
    using layout_type = L;
    using reference = typename layout_type::reference;
    using const_reference = typename layout_type::const_reference;
    
    size_type size() const;
    reference operator[](size_type);
    const_reference operator[](size_type) const;
    
private:
    
    layout_type m_layout;
    array_data m_data;
};
```

Design from @JohanMabille.

JohanMabille:
This API is non exhaustive, the typed_arrya should also have the iterators, front / back, etc ..

PR #65, opened 2024-04-08
Implement typed_array

Alex-PLACET:
Implement #61


include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+    using type =  typename decltype(get_type())::type;
+};
+
+// Template alias for easier usage.
+template<typename T>
+using get_default_layout_t = typename get_default_layout<T>::type;
+
+template <class T, class L = get_default_layout_t<T>>
+class typed_array
```
jjerphan: @JohanMabille mentioned in https://github.com/xtensor-stack/sparrow/issues/61#issuecomment-2039177368 that the API needs to somewhat be similar to `std::vector`, and expose other methods, but I do not know what he has in mind exactly.

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,133 @@
...
+auto typed_array<T, L>::size() const -> size_type
+{
+    return m_layout.size();
+}
+
+template <class T, class L>
+auto typed_array<T, L>::operator[](size_type i) -> reference
+{
+    return m_layout[i];
+}
```
jjerphan: Will this work when `L` is `variable_sized_binary_layout`?

test/test_typed_array.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+#include "doctest/doctest.h"
+
+#include <iostream> // For Doctest
+#include <numeric>
+
+using namespace sparrow;
+
+using data_type_t = int32_t;
+using my_typed_array = typed_array<data_type_t>;
```
jjerphan, line 25: Is there a way to use other types to test all the underlying layouts?

test/test_typed_array.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+namespace
+{
+    array_data make_test_array_data(size_t n = 10, size_t offset = 0)
+    {
+        array_data ad;
+        ad.type = data_descriptor(data_type::INT32);
```
jjerphan, line 32: This can be obtained from the trait via `type_id` now:

https://github.com/xtensor-stack/sparrow/blob/2a07414d2561f078a3f8878623a266608854afaa/include/sparrow/data_traits.hpp#L73

jjerphan (commented):
Thank you, @Alex-PLACET.

Here is a first review.

test/test_typed_array.cpp
Alex-PLACET: I will templatize the tests 👍

JohanMabille (commented):
Thanks for this initial implementation. Here are a few comments, I find it easier to gather them rather than spreading them in the diff:

- Now that #51 is merged, the layout classes' copy and move semantics are deleted. Therefore the `typed_array` class should implement proper semantics.
- Nitpicking: Since the API is going to grow, it could be nice to organize the different methods in the same order as [std::vector](https://en.cppreference.com/w/cpp/container/vector), with the same sections (element access, iterators, etc). Notice that I am following the same "convention" in the incoming refactoring of the buffer class.
- Even if the buffer and the layout classes do not provide all the required methods to implement resize/insert/erase/clear, there are a couple of methods that can already be instantiated: front / back / data / empty (and I'm probably missing some of them)
- I think comparison operators can also be implemented

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,133 @@
...
+    static constexpr auto get_type() {
+        if constexpr (std::is_arithmetic_v<T> || std::is_array_v<T>) {
+            return std::type_identity<sparrow::fixed_size_layout<T>>{};
+        } else {
+            return std::type_identity<sparrow::variable_size_binary_layout<T, T&, const T&>>{};
+        }
+    }
+
+    using type =  typename decltype(get_type())::type;
+};
```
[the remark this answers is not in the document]
Alex-PLACET: Done

include/sparrow/ordering.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <compare>
+#include <concepts>
+
+namespace sparrow
+{
+    template <typename T1, typename T2>
+    concept StrongOrderable = requires(T1 a, T2 b) {
+        { a.value() <=> b.value() } -> std::same_as<std::strong_ordering>;
+    };
```
jjerphan, line 25: How is `StrongOrderable` be picked over `PartialOrderable` when condition on `T1` and `T2` are met?

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+         *
+         * @return The number of elements in the container.
+         */
+        size_type size() const;
+
+        // TODO: Add reserve, capacity, shrink_to_fit
+
+        // Modifiers
+
+        // TODO: Implement insert, erase, push_back, pop_back, clear, resize, swap
```
jjerphan, line 168: Since the diff of this PR is quite large, I guess the following can be implemented in a second PR.

What do you think?

include/sparrow/array_data.hpp
```diff
@@ -282,17 +284,25 @@ namespace sparrow
...
+        if (dlhs && drhs)
+        {
+            return dlhs.value() <=> drhs.value();
+        }
+        return dlhs.has_value() <=> drhs.has_value();
     }
 
     template <class D, not_ref_proxy T>
-    std::strong_ordering operator<=>(const reference_proxy_base<D>& lhs, const T& rhs)
+    std::partial_ordering operator<=>(const reference_proxy_base<D>& lhs, const T& rhs)
```
jjerphan, line 299: Has this been relaxed for floating types?

jjerphan (commented):
A review with some outdated comments, I believe.

include/sparrow/array_data.hpp
Alex-PLACET: Yes it is

include/sparrow/typed_array.hpp
Alex-PLACET: Yes, and currently there is no way to do mutable operations on the arrays

include/sparrow/algorithm.hpp
```diff
@@ -0,0 +1,104 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <algorithm>
+#include <compare>
+
+#if defined(__APPLE__) && defined(__clang__) && defined(__APPLE_CC__)
```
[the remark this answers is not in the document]
jjerphan: Resolving the thread as it is outdated.

include/sparrow/ordering.hpp
Alex-PLACET: It's about the returned type pf the spaceship operator

test/test_typed_array.cpp
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        SUBCASE("at")
+        {
+            const auto array_data = make_test_array_data<T>(10, 1);
+            typed_array<T> typed_array(array_data);
+            for (typename ::typed_array<T>::size_type i = 0; i < typed_array.size() - 1; ++i)
+            {
+                const T value = to_value_type<T>(i + 1);
+                CHECK_EQ(typed_array.at(i).value(), value);
```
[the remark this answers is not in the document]
Alex-PLACET: It was for debug purpose, I will change that

include/sparrow/details/3rdparty/float16_t.hpp
```diff
@@ -786,6 +788,15 @@ namespace numeric
         return !( lhs == rhs );
     }
 
+    // Added by Alexis Placet from Quantstack
```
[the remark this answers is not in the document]
Alex-PLACET: I added a note at the top of the file

jjerphan (approved):

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+        /**
+         * @brief Returns an iterator to the end.
+         *
+         * @description Returns an iterator to the element following the last element of the vector.
+         * This element acts as a placeholder; attempting to access it results in undefined behavior.
+         *
+         * @return An iterator to the end.
+         */
+        iterator end();
+        const_iterator cend() const;
```
JohanMabille, line 127: the const overload of `end` is missing

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+        /*
+         * @brief Returns an iterator to the beginning.
+         *
+         * @description Returns an iterator to the first element of the vector.
+         * If the vector is empty, the returned iterator will be equal to end().
+         *
+         * @return An iterator to the beginning.
+         */
+        iterator begin();
+        const_iterator cbegin() const;
```
JohanMabille, line 114: The const overload of `begin` is missing (you generally need it in generic code that accepts both const and non const objects).

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+        // Modifiers
+
+        // TODO: Implement insert, erase, push_back, pop_back, clear, resize, swap
+
+        std::partial_ordering operator<=>(const typed_array<T>& other) const;
+
+        bool operator==(const typed_array<T>& other) const;
+
+        bool operator!=(const typed_array<T>& other) const;
```
JohanMabille, line 174: In C++20, this operator is automatically generated when you implement `operator==`

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+        bool operator<(const typed_array<T>& other) const;
+
+        bool operator>(const typed_array<T>& other) const;
+
+        bool operator<=(const typed_array<T>& other) const;
+
+        bool operator>=(const typed_array<T>& other) const;
```
JohanMabille, line 182: In C++20, these four operators are automatically generated when you implement `operator<=>`

include/sparrow/ordering.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#pragma once
+
+#include <compare>
+#include <concepts>
+
+namespace sparrow
+{
+    template <typename T1, typename T2>
+    concept StrongOrderable = requires(T1 a, T2 b) {
```
JohanMabille, line 23: Naming convention: we prefer snake_case over CamelCase.

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+                + " at index " + std::to_string(i)
+            );
+        }
+        return m_layout[i];
+    }
+
+    template <class T, class L>
+    auto typed_array<T, L>::operator[](size_type i) -> reference
+    {
+        return m_layout[i];
```
JohanMabille, line 229: Consider adding an assert here and in the following methods.

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+         */
+        size_type size() const;
+
+        // TODO: Add reserve, capacity, shrink_to_fit
+
+        // Modifiers
+
+        // TODO: Implement insert, erase, push_back, pop_back, clear, resize, swap
+
+        std::partial_ordering operator<=>(const typed_array<T>& other) const;
```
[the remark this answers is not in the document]
Klaim: Looks like all the comparision operators can be defaulted?

include/sparrow/typed_array.hpp
Klaim: Yes, set the asserts matching the preconditions and type-invariants 👍🏽 .

test/array_data_creation.hpp
```diff
@@ -0,0 +1,89 @@
...
+        return ad;
+    }
+
+    template <>
+    sparrow::array_data make_test_array_data<std::string>(size_t n, size_t offset)
+    {
+        std::vector<std::string> words;
+        for (size_t i = 0; i < n; ++i)
+        {
+            words.push_back(std::to_string(i));
```
Klaim, line 52: We might want to make the genreation more random in the future haha

test/test_typed_array.cpp
```diff
@@ -0,0 +1,290 @@
...
+            const auto array_data = make_test_array_data<T>(10, 1);
+            const typed_array<T> typed_array(array_data);
+            CHECK_EQ(typed_array.size(), 9);
+        }
+
+        SUBCASE("at")
+        {
+            const auto array_data = make_test_array_data<T>(10, 1);
+            typed_array<T> typed_array(array_data);
+            for (typename ::typed_array<T>::size_type i = 0; i < typed_array.size() - 1; ++i)
```
Klaim, line 64: why does `::typed_array`  the `::`  is necessary here and in other tests below?

Alex-PLACET: The issue comes from the declaration of the variable which have the same name as the type. I will change that

test/array_data_creation.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace
+{
+    template <typename T>
+    sparrow::array_data make_test_array_data(size_t n = 10, size_t offset = 0)
+    {
+        sparrow::array_data ad;
+        ad.type = sparrow::data_descriptor(sparrow::arrow_traits<T>::type_id);
+        ad.bitmap = sparrow::dynamic_bitset<uint8_t>(n, true);
+        if (n >= 10)
```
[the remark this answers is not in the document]
Alex-PLACET: You right it's unclear, I will improve the function to let the user set the bitmap

test/test_typed_array.cpp
```diff
@@ -0,0 +1,290 @@
...
+            CHECK_EQ(typed_array_empty.cbegin(), typed_array_empty.cend());
+        }
+
+        SUBCASE("bitmap")
+        {
+            const auto array_data = make_test_array_data<T>(10, 1);
+            const typed_array<T> typed_array(array_data);
+            const auto bitmap = typed_array.bitmap();
+            CHECK_EQ(bitmap.size(), 9);
+            for (size_t i = 0; i < bitmap.size() - 1; ++i)
```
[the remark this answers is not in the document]
Alex-PLACET: The bitmap range taks into account the offset. As the the offset is 1, the value of the bitmap range at 9 - 1 (8) is false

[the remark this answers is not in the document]
Alex-PLACET: I will replace all the magic numbers with constexpr to have code more expressive

include/sparrow/details/3rdparty/float16_t.hpp
```diff
@@ -3,17 +3,21 @@
...
 //
 // inspired by:
 // https://github.com/acgessler/half_float
 // https://github.com/x448/float16
 //
+#include <compare>
 #include <cstdint>
 #include <limits>
 #include <iostream>
+#include <sstream>
```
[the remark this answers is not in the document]
Alex-PLACET: Without, I get this error: 
```
float16_t.hpp:804:49: error: implicit instantiation of undefined template 'std::basic_ostringstream<char>'
  804 |         std::basic_ostringstream<CharT, Traits> __s;
      |                                                 ^
doctest/doctest.h:1155:9: note: in instantiation of function template specialization 'numeric::operator<<<char, std::char_traits<char>>' requested here
 1155 |         operator<<(*stream, in);
      |         ^
doctest/doctest.h:1053:22: note: in instantiation of member function 'doctest::detail::filldata<numeric::float16_t>::fill' requested here
 1053 |         filldata<T>::fill(stream, in);
```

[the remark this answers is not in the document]
Alex-PLACET: I moved the include in the test file 👍

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+    // Element access
+
+    template <class T, class L>
+    auto typed_array<T, L>::at(size_type i) -> reference
+    {
+        if (i >= size())
+        {
+            throw std::out_of_range(
+                "typed_array::at: index out of range for array of size " + std::to_string(size())
```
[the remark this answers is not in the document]
Alex-PLACET: <format> is not available on Apple Clang. Do you want a specific code  in that case ?

include/sparrow/ordering.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+    template <typename T1, typename T2>
+        requires PartialOrderable<T1, T2>
+    struct OrderingType<T1, T2>
+    {
+        using type = std::partial_ordering;
+    };
+
+    template <typename T1, typename T2>
+    using OrderingType_t = typename OrderingType<T1, T2>::type;
```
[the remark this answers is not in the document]
Alex-PLACET: I simplified the code by using and I don't need it anymore

include/sparrow/ordering.hpp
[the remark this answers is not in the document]
Alex-PLACET: It was removed

include/sparrow/algorithm.hpp
```diff
@@ -0,0 +1,100 @@
...
+namespace sparrow
+{
+    template <class R1, class R2, class Cmp>
+    constexpr auto lexicographical_compare_three_way(const R1& r1, const R2& r2, Cmp comp)
+        -> decltype(comp(*r1.cbegin(), *r2.cbegin()))
+    {
+#if !COMPILING_WITH_APPLE_CLANG
+        return std::lexicographical_compare_three_way(r1.cbegin(), r1.cend(), r2.cbegin(), r2.cend(), comp);
+#else
+        auto f1 = r1.cbegin();
```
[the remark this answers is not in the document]
Alex-PLACET: I finally changed the code to have something more readable. I also used a concept instead of the assert

test/test_typed_array.cpp
```diff
@@ -0,0 +1,290 @@
...
+        }
+    }
+}
+
+TEST_SUITE("typed_array")
+{
+    TEST_CASE_TEMPLATE_DEFINE("scalar", T, scalar)
+    {
+        // Element access
+
```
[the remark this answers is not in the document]
Alex-PLACET: Will be done in another PR

include/sparrow/config.hpp
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#if defined(__apple_build_version__)
+#define COMPILING_WITH_APPLE_CLANG
+#endif
+
+consteval bool is_apple_compiler()
```
jjerphan, line 21: Would it makes sense to have it `constinit`?

jjerphan (commented):
Definitely LGTM.

I just have one comment.

include/sparrow/typed_array.hpp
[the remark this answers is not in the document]
Klaim: Apple Clang is starting to be annoying T_T
OK nevermind, keep this code and add a `// TODO: use <format> instead when Apple Clang provideds it ` in all functions where you concatenate strings 👍🏽 We'll see that later.

test/test_typed_array.cpp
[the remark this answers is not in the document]
Klaim: (note for future reader: discussed in chat, maybe we wont need it, we'll see)

include/sparrow/config.hpp
Klaim: `constinit` is only for objects and is forcing the initial value to be evaluated at compile-time. Here `consteval` means this fonction can only be called at compile-time (therefore, in "constant expressions" that is code which runs at compile-time). If it was an object instead of a function then  `constexpr` (defined at compile time, can be accessed at runtime) would be best, as `consteval` (exists only at compile-time) is not (yet) allowed on anything other than functions.

.github/workflows/windows.yml
```diff
@@ -41,6 +41,26 @@ jobs:
         echo "CC=clang" >> $GITHUB_ENV
         echo "CXX=clang++" >> $GITHUB_ENV
 
+    - name: Setup MinGW
+      if: matrix.sys.compiler == 'default'
```
Klaim, line 45: Note that this is not complete enough: mingw is used only if Ninja is the build-system used through cmake, if the compiler is set to `default` and the build-system too, then it will be msvc + msbuild, which will not use this install. Maybe just add the condition for ninja?
Same for below setup which if my yunderstanding is correct will force mingw even when msbuild is used

include/sparrow/algorithm.hpp
```diff
@@ -0,0 +1,126 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <algorithm>
+#include <compare>
+
```
Klaim, line 19:
```suggestion
#include <concepts>
```

include/sparrow/config.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+#if defined(__apple_build_version__)
```
Klaim, line 23: Suggestion: maybe make `COMPILING_WITH_APPLE_CLANG` be `1` or `0` and then just return here that casted to `bool`? That way we check specifically once if we are using apple clang, but then use the result in the define and here.

include/sparrow/data_traits.hpp
```diff
@@ -114,15 +114,15 @@ namespace sparrow
     {
         static constexpr data_type type_id = data_type::STRING;
         using value_type = std::string;
-        using default_layout = variable_size_binary_layout<value_type, std::string_view, std::string_view>;  // FIXME: this is incorrect, change when we have the right types
+        using default_layout = variable_size_binary_layout<value_type, std::string_view, const std::string_view>;  // FIXME: this is incorrect, change when we have the right types
```
Klaim, line 117: Is the FIXME comment still true or is this the final type for the read-only view?
(same question below)

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,360 @@
...
+
+#if !defined(COMPILING_WITH_APPLE_CLANG)
+#include <format>
+#endif
+
+namespace sparrow
+{
+    template <class T, class L = typename arrow_traits<T>::default_layout>
+        requires is_arrow_base_type<T>
+    class typed_array
```
Klaim, line 33: The documentation for this type is missing (I mean the overview of the type, comments before it's definition)

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,360 @@
...
+         *
+         * @return An iterator to the element following the last element of the vector.
+         */
+        iterator end();
+        const_iterator end() const;
+        const_iterator cend() const;
+        ///@}
+
+        /*
+         * @return A range of the bitmap.
```
Klaim, line 134:
```suggestion
         * @return A range of the bitmap. For each index position in this range, if `true` then there is a value at the same index position in the `values()` range, `false` means the value there is null.
```

include/sparrow/typed_array.hpp
```diff
@@ -0,0 +1,360 @@
...
+        return m_layout[i];
+    }
+
+    template <class T, class L>
+        requires is_arrow_base_type<T>
+    auto typed_array<T, L>::at(size_type i) const -> const_reference
+    {
+        if (i >= size())
+        {
+#if defined(COMPILING_WITH_APPLE_CLANG)
```
Klaim, line 208: We might want to make our own formatting function that forwards to std::format or  custom impl based on std::stringstream. Can wait another PR of course.

test/array_data_creation.hpp
```diff
@@ -0,0 +1,123 @@
...
+#pragma once
+
+#include <numeric>
+
+#include "sparrow/array_data.hpp"
+#include "sparrow/data_traits.hpp"
+
+namespace sparrow::test
+{
+    /**
```
Klaim, line 24: Use normal comments in tests, as these functions will not be part of the documented api. (we dont want this to appear in the documentation)

test/test_typed_array.cpp
```diff
@@ -0,0 +1,309 @@
...
+TEST_SUITE("typed_array")
+{
+    TEST_CASE_TEMPLATE_DEFINE("all", T, all)
+    {
+        SUBCASE("constructor with parameter")
+        {
+            constexpr size_t n = 10;
+            constexpr size_t offset = 1;
+            const auto array_data = sparrow::test::make_test_array_data<T>(n, offset);
+            const typed_array<T> typed_array(array_data);
```
Klaim, line 63: Consider using `ta{array_data}` instead of parenthesis so that we benefit from no-implicit-conversion-allowed (same for all functions below). I suspect we will not provide initializer list at the typed_array level, but if you think we will then ignore my remark.

test/test_variable_size_binary_layout.cpp
```diff
@@ -33,7 +33,7 @@ namespace sparrow
             m_data.buffers[1].resize(std::accumulate(
                 words,
                 words + nb_words,
-                0u,
+                size_t(0),
```
Klaim, line 36: note: c++23 adds the uz familly of litterals, which would have been appropriate here 🥲

include/sparrow/data_traits.hpp
Alex-PLACET: I don't know. I had to fix this to pass my UT. Where is the discussion about that topic ?
@JohanMabille any idea ?

test/array_data_creation.hpp
Alex-PLACET: Can't we exclude the test folder from the parsing ?

include/sparrow/typed_array.hpp
Alex-PLACET: Ok, I will keep the std::to_string implementation and wait for our custom format implemenation. I'm struggle with the GCC support on Windows and it's not the point of this PR.

.github/workflows/windows.yml
Alex-PLACET: I will remove the update to GCC 13 for the moment as it's not the point of this PR and it's more difficult than expected and because there no need in this PR as I removed the use of <format>

include/sparrow/data_traits.hpp
Klaim: There: https://github.com/xtensor-stack/sparrow/pull/49/files#r1542484372

test/array_data_creation.hpp
Klaim: No idea but I suppose yes, but even if we do the doxygen stuffs is both misleading (as in suggests it should appear in the some doc or at least is "official api") and unnecessarilly noisy.

include/sparrow/data_traits.hpp
Alex-PLACET: Thanks for this link. So let's keep the comment to change the type later.

include/sparrow/typed_array.hpp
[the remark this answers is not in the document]
Alex-PLACET: I added a comment and kept the code with std::to_string

JohanMabille (approved):

SylvainCorlay:
Congrats on getting the PR in @Alex-PLACET !

PR #70, opened 2024-04-09
Support for timestamps

### LINKED ISSUE #52: Support `timestamps`
For Arrow, see:
 - [the section of the Apache Arrow documentation on Timestamps](https://arrow.apache.org/docs/python/timestamps.html)
 - [`TimestampType` code comment](https://github.com/apache/arrow/blob/83359d6958273be534f376f40976196c14675c8b/cpp/src/arrow/type.h#L1632-L1663)

For C++20, see:
 - [`std::chrono::time_zone`](https://en.cppreference.com/w/cpp/chrono/time_zone)
 - [`std::chrono::zoned_time`](https://en.cppreference.com/w/cpp/chrono/zoned_time)

jjerphan:
Fix #52.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,15 +13,16 @@ defaults:
 jobs:
   build:
     runs-on: ${{ matrix.runs-on }}
-    name: ${{ matrix.sys.compiler }}-${{ matrix.build-system }}-${{ matrix.config.name }}
+    name: ${{ matrix.sys.compiler }} / ${{ matrix.build-system }} / ${{ matrix.config.name }} / date-polyfill ${{ matrix.sys.date-polyfill}}
     strategy:
       fail-fast: false
       matrix:
         runs-on: [windows-latest]
         sys:
-        - {compiler: default}
-        - {compiler: msvc}
-        - {compiler: clang}
+        - {compiler: msvc, date-polyfill: 'ON' }
+        - {compiler: msvc, date-polyfill: 'OFF' }
+        - {compiler: clang, date-polyfill: 'ON' }
+        - {compiler: clang, date-polyfill: 'OFF' }
         config:
         - { name: Debug }
         - { name: Release }
@@ -55,7 +56,7 @@ jobs:
           ninja
 
     - name: Configure using CMake
-      run: cmake -Bbuild -DCMAKE_BUILD_TYPE:STRING=${{matrix.config.name}} -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DBUILD_TESTS=ON -G "${{matrix.build-system}}"
+      run: cmake -Bbuild -DCMAKE_BUILD_TYPE:STRING=${{matrix.config.name}} -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DBUILD_TESTS=ON -DUSE_DATE_POLYFILL=${{matrix.sys.date-polyfill}} -G "${{matrix.build-system}}"
 
     - name: Install
       working-directory: build
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -19,10 +19,10 @@ jobs:
       matrix:
         runs-on: [windows-latest]
         sys:
-        - {compiler: msvc, date-polyfill: 'ON' }
-        - {compiler: msvc, date-polyfill: 'OFF' }
-        - {compiler: clang, date-polyfill: 'ON' }
-        - {compiler: clang, date-polyfill: 'OFF' }
+        - { compiler: msvc, date-polyfill: 'ON' }
+        - { compiler: msvc, date-polyfill: 'OFF' }
+        - { compiler: clang, date-polyfill: 'ON', version: 17 }
+        - { compiler: clang, date-polyfill: 'OFF', version: 17 }
         config:
         - { name: Debug }
         - { name: Release }
@@ -35,6 +35,13 @@ jobs:
     - name: Setup MSVC
       if: matrix.sys.compiler == 'msvc' && matrix.build-system != 'Visual Studio 17 2022'
       uses: ilammy/msvc-dev-cmd@v1
+
+    - name: Install LLVM and Clang
+      if: matrix.sys.compiler == 'clang'
+      uses: egor-tensin/setup-clang@v1
+      with:
+        version: ${{matrix.sys.version}}
+        platform: x64
 
     - name: Setup clang
       if: matrix.sys.compiler == 'clang'
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -7,3 +7,8 @@ dependencies:
   - ninja
   # Tests
   - doctest
+  # P0355R7 (Extending chrono to Calendars and Time Zones) has not been entirely implemented in libc++ yet.
+  # See: https://libcxx.llvm.org/Status/Cxx20.html#note-p0355
+  # For now, we use HowardHinnant/date as a replacement if we are compiling with libc++.
+  # TODO: remove this once libc++ has full support for P0355R7.
+  - howardhinnant_date
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -5,5 +5,7 @@ dependencies:
   # Build
   - cmake
   - ninja
+  # Alternative of `std::chrono` introduced in C++20.
+  - howardhinnant_date
   # Tests
   - doctest
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -5,7 +5,5 @@ dependencies:
   # Build
   - cmake
   - ninja
-  # Alternative of `std::chrono` introduced in C++20.
-  - howardhinnant_date
   # Tests
   - doctest
--- a/include/sparrow/data_traits.hpp
+++ b/include/sparrow/data_traits.hpp
@@ -125,6 +125,12 @@ namespace sparrow
         using default_layout = variable_size_binary_layout<value_type, std::span<byte_t>, const std::span<byte_t>>;  // FIXME: this is incorrect, change when we have the right types
     };
 
+    template <>
+    struct arrow_traits<timestamp> : common_native_types_traits<timestamp>
+    {
+        static constexpr data_type type_id = data_type::TIMESTAMP;
+    };
+
     namespace predicate
     {
 
--- a/include/sparrow/data_type.hpp
+++ b/include/sparrow/data_type.hpp
@@ -17,11 +17,11 @@
 #include <version>
 #include <chrono>
 
-// P0355R7 (Extending chrono to Calendars and Time Zones) has not been entirely implemented in libc++ yet.
-// See: https://libcxx.llvm.org/Status/Cxx20.html#note-p0355
-// For now, we use HowardHinnant/date as a replacement if we are compiling with libc++.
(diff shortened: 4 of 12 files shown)


include/sparrow/data_traits.hpp
```diff
@@ -18,6 +18,8 @@
 #include "fixed_size_layout.hpp"
 #include "variable_size_binary_layout.hpp"
 
+using time_point = std::chrono::time_point<std::chrono::system_clock>;
```
Alex-PLACET, line 21: Do you really want to do it outside the sparrow namespace ?

jjerphan: Indeed, we probably should have it part of the `sparrow` namespace.

.github/workflows/windows.yml
```diff
@@ -27,7 +27,9 @@ jobs:
         - { name: Release }
         build-system:
         - "Visual Studio 17 2022"
-        - "Ninja"
+        # Ninja is disabled because the standard library implementation is uses does not
```
Alex-PLACET, line 30: Build system does not implie the use of a specific standard library.
When you use Ninja as generator you have to define the compilers and toolset in case you want MSVC:
cmake -G "Ninja" -T "v142" -DCMAKE_C_COMPILER="C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe" -DCMAKE_CXX_COMPILER="C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe"

jjerphan: I should have worded this comment differently, indeed.

I am not knowledgeable of Windows' toolchains at all. Feel free to adapt it and the toolchain as you wish.

Klaim: See the `Setup clang` step for how clang is used with both ninja and msbuild.

Klaim: > I am not knowledgeable of Windows' toolchains at all.

To be short: 
- Windows as an ABI which is usable with libstd-c++, libc++, and msvc-stl , which are respectively provided by the *build toolchains* of mingw, llvm/clang and msvc (cl is the compiler, msvc is the toolchain). 
- each *build toolchain* has a prefered stl/runtime implementation by default, it's changed only if specified explicitly: 
    -gcc/mingw will use libstd-c++ by default
    - clang will use msvc-stl by default (even if libc++ is installed with it, it has to be requested explicitly to be used)
    - msvc can only use msvc-stl
- the *build system* is irrelevant to the choice of the build toolchain, however they have different defaults if you dont specificy the build-toolchain:
    - msbuild (generator "Visual Studio blah") will use msvc by default, therefore msvc-stl
    - ninja will use mingw by default, therefore libstdc++
- finally, cmake also has a default generator which depends on the development platform:
    - on windows, if not specified, it will use the "Visual Studio ..." generator with everything by default
    - on linux it will use the "Make" generator by default
    - not sure on MacOSX but I believe it generates something known by XCode by default
    - AFAIK Ninja is never a default generator
    - using either ninja or make or similar agnostic build-systems implies probably to also specify which compiler (and therfore build toolchain and therefore stl and runtime impl) to use

jjerphan:
[P0355](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0355r7.html) has not [been implemented entirely by `libc++`](https://libcxx.llvm.org/Status/Cxx20.html#note-p0355) which _has_ to be used by clang on conda-forge for ABI compatibility and for runtime uniqueness.

The plan is vendor parts of [`HowardHinnant/date`](https://github.com/HowardHinnant/date) and use it when `-stdlib=libc++`.

jjerphan:
See https://github.com/conda-forge/staged-recipes/pull/26073

environment-dev.yml
```diff
@@ -5,5 +5,7 @@ dependencies:
   # Build
   - cmake
   - ninja
+  # Alternative of `std::chrono` introduced in C++20.
+  - howardhinnant_date
```
JohanMabille, line 9: Since this is required for LIBCPP only, I think it should not be in the general environment file.

JohanMabille (approved):
I don't think we need additional tests to check compilation since `all_base_types_t` is already instantiated in the metaprogramming tests.

environment-dev.yml
jjerphan: Yes, I wish selectors were present (see https://github.com/conda/conda/issues/8089).

Let me modify the CI accordingly.

jjerphan:
For reference, feature testing macros aren't present in stdlibc++ 12 and 13. I have not checked for stdlibc++14 yet, but we can't expect to use it, I believe.

include/sparrow/data_type.hpp
```diff
@@ -14,6 +14,16 @@
 
 #pragma once
 
+#include <version>
+#include <chrono>
+
+#if __cpp_lib_chrono > 201907L
```
jjerphan, line 20: `__cpp_lib_chrono` (like any other FTM) might not be implemented in compiler, and it thus unusable.

See the discussions in https://github.com/llvm/llvm-project/issues/73953.

jjerphan:
As discussed, let's use `HowardHinnant/date` everywhere until all the implementations of the C++20 standard libraries are available.

jjerphan:
Failures on the CI are being addressed by #89 and #92 respectively.

Alex-PLACET (changes requested):
You should add the type in the TEST_CASE_TEMPLATE_INVOKE of test_typed_array.cpp

jjerphan:
In fact, dedicated tests for `timestamps` are required because floating or integral types are cannot be converted to them and because their set is not strongly ordered as well.

include/sparrow/data_type.hpp
```diff
@@ -103,7 +120,8 @@ namespace sparrow
         float32_t,
         float64_t,
         std::string,
-        std::vector<byte_t>
+        std::vector<byte_t>,
+        sparrow::timestamp
```
Klaim, line 124: I think there is one type missing before that one, see compare with the order of the enum, `BINARY` matches `std::vector<byte_t>` but we dont have something for `FIXED_SIZE_BINARY`. Not sure which type is missing though. @JohanMabille any idea what it would match?

jjerphan: I think `FIXED_SIZE_BINARY` is meant to be used for a fixed size binary type which is not a primitive type, in which case we could use a `struct` as a type to test `FIXED_SIZE_BINARY`.

JohanMabille: I took the values from [Arrow implementation](https://github.com/apache/arrow/blob/main/cpp/src/arrow/type_fwd.h), not sure which type this value should match.

jjerphan: Also, must we work on supporting [`DATE32` and `DATE64` which come before `TIMESTAMP`](https://github.com/apache/arrow/blame/b2e8c33c86c819b167a1cbca834da3c9047a9350/cpp/src/arrow/type_fwd.h#L358-L364)?

JohanMabille: The support for DATE32 and DATE64 can be added later. Explicit integer identifiers would be required later for backward compatibility, but it doe snot matter for now, nor the enum "stability".

jjerphan:
Failures are related to https://github.com/HowardHinnant/date/issues/788.

jjerphan:
https://github.com/conda-forge/howardhinnant_date-feedstock/pull/10 addresses it for conda-forge.

We might need to have a fix for vcpkg as well.

Unrelated but informative: https://github.com/conan-io/conan-center-index/pull/21671/ addresses it for Conan.

SylvainCorlay:
I iterated on the PR

 - added a cmake flag `-D USE_DATE_POLYFILL` defaulting to `ON`, determining whether to use `howardhinnant_date` or `std::chrono`.
 - added tests to the matrix for testing with both `howardhinnant_date` and `std::chrono` on platforms where it is supported.

I removed the mingw tests because
 - there is no MinGW build of `howardhinnant_date`.
 - the GCC version used by MinGW is 12, which does not support the required features yet. We will be able to re-add this when the windows runners include GCC 13.

On a separate note, a difference between the polyfill and std::chrono is that the literals in the polyfill are prefixed with `_` so I removed the use of the literals in the tests.

.github/workflows/windows.yml
```diff
@@ -13,15 +13,16 @@ defaults:
...
+    name: ${{ matrix.sys.compiler }} / ${{ matrix.build-system }} / ${{ matrix.config.name }} / date-polyfill ${{ matrix.sys.date-polyfill}}
     strategy:
       fail-fast: false
       matrix:
         runs-on: [windows-latest]
         sys:
-        - {compiler: default}
-        - {compiler: msvc}
-        - {compiler: clang}
+        - {compiler: msvc, date-polyfill: 'ON' }
```
jjerphan, line 22: FYI, I am subscribed to GitHub Action Runners' new releases.
```suggestion
        # TODO: Reintroduce default (or just mingw) when the new version is release
        # See: https://github.com/actions/runner-images/commits/main/images/windows/Windows2022-Readme.md
        # See: https://www.mingw-w64.org/changelog/
        - {compiler: msvc, date-polyfill: 'ON' }
```

jjerphan (commented):
Thank you, Sylvain. I just have one suggestion before merging this PR.

For the readers' information, those issues are worth reading and watching:
 - https://github.com/HowardHinnant/date/issues/828
 - https://github.com/HowardHinnant/date/issues/565

.github/workflows/windows.yml
SylvainCorlay: I suggest we explicitely invoke mingw rather than calling it because it is the default toolchain with Ninja.

test/test_typed_array_timestamp.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <iostream>  // For Doctest
+#include <sstream>
+#include <string>
+#include <type_traits>
+
+#if defined(SPARROW_USE_DATE_POLYFILL)
+#include <date/tz.h>
+#else
```
Alex-PLACET, line 26: Already done in sparrow/data_type.hpp

test/test_typed_array_timestamp.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    // Element access
+
+    TEST_CASE("at")
+    {
+        const auto array_data = sparrow::test::make_test_array_data<sparrow::timestamp>(test_n, test_offset, false_bitmap);
+        sparrow::typed_array<sparrow::timestamp> ta{array_data};
+        for (typename sparrow::typed_array<sparrow::timestamp>::size_type i = 0; i < ta.size() - 1; ++i)
+        {
+            CHECK_EQ(ta.at(i).value(), sparrow::timestamp(date::sys_days(date::year(1970)/date::January/date::day(1)) + date::days(i + 1)));
```
Alex-PLACET, line 61: Can you extract `date::sys_days(date::year(1970)/date::January/date::day(1))` in a constant to ease the readibility ?

Alex-PLACET (approved):

JohanMabille (approved):

PR #74, opened 2024-04-10
Refactored buffer and buffer_view

### LINKED ISSUE #45: Refactor buffer internal representation
The buffer class should use 3 pointers instead of pointer + size (+ capacity), as the implementation of `std::vector` does.
This should improve performance and make it easier to implement the insert / reserve / push_back methods, as well as the noexcept handling.

JohanMabille:
Fix #45

### DIFF
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -63,7 +63,7 @@ namespace sparrow
         constexpr buffer_base(const A& a) noexcept;
 
         template <allocator A = allocator_type>
-        buffer_base(size_type n, const A& a = A());
+        constexpr buffer_base(size_type n, const A& a = A());
 
         template <allocator A = allocator_type>
         constexpr buffer_base(pointer p, size_type n, const A& a = A());
@@ -80,9 +80,9 @@ namespace sparrow
         constexpr buffer_data& get_data() noexcept;
         constexpr const buffer_data& get_data() const noexcept;
 
-        pointer allocate(size_type n);
-        void deallocate(pointer p, size_type n);
-        void create_storage(size_type n);
+        constexpr pointer allocate(size_type n);
+        constexpr void deallocate(pointer p, size_type n);
+        constexpr void create_storage(size_type n);
         constexpr void assign_storage(pointer p, size_type n, size_type cap);
 
     private:
@@ -300,7 +300,7 @@ namespace sparrow
     
     template <class T>
     template <allocator A>
-    buffer_base<T>::buffer_base(size_type n, const A& a)
+    constexpr buffer_base<T>::buffer_base(size_type n, const A& a)
         : m_alloc(a)
     {
         create_storage(n);
@@ -353,19 +353,19 @@ namespace sparrow
     }
 
     template <class T>
-    auto buffer_base<T>::allocate(size_type n) -> pointer
+    constexpr auto buffer_base<T>::allocate(size_type n) -> pointer
     {
         return alloc_traits::allocate(m_alloc, n);
     }
 
     template <class T>
-    void buffer_base<T>::deallocate(pointer p, size_type n)
+    constexpr void buffer_base<T>::deallocate(pointer p, size_type n)
     {
         alloc_traits::deallocate(m_alloc, p, n);
     }
 
     template <class T>
-    void buffer_base<T>::create_storage(size_type n)
+    constexpr void buffer_base<T>::create_storage(size_type n)
     {
         m_data.p_begin = allocate(n);
         m_data.p_end = m_data.p_begin + n;
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -95,10 +95,10 @@ namespace sparrow
      * @class buffer
      * @brief Object that owns a piece of contiguous memory
      *
-     * This class provides an API similat to std::vector, with
+     * This class provides an API similar to std::vector, with
      * two main differences:
-     * - it is not templated by the allocator type, but type-erases
-     *   it instead.
+     * - it is not templated by the allocator type, but makes use of
+     *   any_allocator which type-erases it.
      * - it can acquire ownership of an already allocated raw buffer.
      */
     template <class T>
@@ -134,19 +134,19 @@ namespace sparrow
         }
 
         template <allocator A = allocator_type>
-        explicit buffer(size_type n, const A& a = A());
+        constexpr explicit buffer(size_type n, const A& a = A());
 
         template <allocator A = allocator_type>
-        buffer(size_type n, const value_type& v, const A& a = A());
+        constexpr buffer(size_type n, const value_type& v, const A& a = A());
 
         template <allocator A = allocator_type>
-        buffer(pointer p, size_type n, const A& a = A());
+        constexpr buffer(pointer p, size_type n, const A& a = A());
 
         template <allocator A = allocator_type>
-        buffer(std::initializer_list<value_type> init, const A& a = A());
+        constexpr buffer(std::initializer_list<value_type> init, const A& a = A());
 
         template <class It, allocator A = allocator_type>
-        buffer(It first, It last, const A& a = A());
+        constexpr buffer(It first, It last, const A& a = A());
 
         ~buffer();
 
@@ -267,7 +267,6 @@ namespace sparrow
     template <class T>
     constexpr bool operator==(const buffer<T>& lhs, const buffer<T>& rhs) noexcept;
 
-
     /******************************
      * buffer_base implementation *
      ******************************/
@@ -376,6 +375,7 @@ namespace sparrow
     template <class T>
     constexpr void buffer_base<T>::assign_storage(pointer p, size_type n, size_type cap)
     {
+        assert(n <= cap);
         m_data.p_begin = p;
         m_data.p_end = p + n;
         m_data.p_storage_end = p + cap;
@@ -387,7 +387,7 @@ namespace sparrow
 
     template <class T>
     template <allocator A>
-    buffer<T>::buffer(size_type n, const A& a)
+    constexpr buffer<T>::buffer(size_type n, const A& a)
         : base_type(check_init_length(n, a), a)
     {
         get_data().p_end = default_initialize(get_data().p_begin, n, get_allocator());
@@ -395,7 +395,7 @@ namespace sparrow
 
     template <class T>
     template <allocator A>
-    buffer<T>::buffer(size_type n, const value_type& v, const A& a)
+    constexpr buffer<T>::buffer(size_type n, const value_type& v, const A& a)
         : base_type(check_init_length(n, a), a)
     {
         get_data().p_end = fill_initialize(get_data().p_begin, n, v, get_allocator());
@@ -403,14 +403,14 @@ namespace sparrow
 
     template <class T>
     template <allocator A>
-    buffer<T>::buffer(pointer p, size_type n, const A& a)
+    constexpr buffer<T>::buffer(pointer p, size_type n, const A& a)
         : base_type(p, check_init_length(n, a), a)
     {
     }
 
     template <class T>
     template <allocator A>
-    buffer<T>::buffer(std::initializer_list<value_type> init, const A& a)
+    constexpr buffer<T>::buffer(std::initializer_list<value_type> init, const A& a)
         : base_type(check_init_length(init.size(), a), a)
     {
         get_data().p_end = copy_initialize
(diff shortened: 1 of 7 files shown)


include/sparrow/buffer.hpp
```diff
@@ -18,454 +18,870 @@
...
     }
 
     template <class T>
-    bool buffer_base<T>::empty() const noexcept
+    buffer<T>& buffer<T>::operator=(const buffer& rhs)
     {
-        return size() == size_type(0);
+        if (std::addressof(rhs) != this)
+        {
+            // We assume that any_allocator nevers propagates on assign
```
jjerphan, line 500: What do you mean by this comment exactly?
```suggestion
            // We assume that any_allocator never propagates on assign.
```

include/sparrow/buffer.hpp
```diff
@@ -18,454 +18,870 @@
...
     }
 
     template <class T>
-    buffer<T>::buffer(pointer data, size_type size)
-        : base_type{data, size}
+    constexpr void buffer<T>::reserve(size_type new_cap)
     {
+        if (new_cap > max_size())
+        {
+            throw std::length_error("buffer::reserve called with new_cap > max_size()");
```
jjerphan, line 697: What should the policy for raising exception be?

include/sparrow/buffer.hpp
```diff
@@ -18,454 +18,870 @@
...
     }
 
     template <class T>
-    void buffer<T>::clear()
+    constexpr auto buffer<T>::max_size_impl(const allocator_type& a) noexcept -> size_type
     {
-        resize(size_type(0));
+        const size_type diff_max = std::numeric_limits<difference_type>::max();
+        const size_type alloc_max = std::allocator_traits<allocator_type>::max_size(a);
+        return (std::min)(diff_max, alloc_max);
```
jjerphan, line 836: Why is this needed?
```suggestion
        return std::min(diff_max, alloc_max);
```

jjerphan (commented):
A first brief pass.

include/sparrow/buffer.hpp
JohanMabille: We do not assign allocators from a buffer instance to another one.

include/sparrow/buffer.hpp
JohanMabille: I took the same as `std::vector`, see the `Exceptions` section [here](https://en.cppreference.com/w/cpp/container/vector/vector) and [here](https://en.cppreference.com/w/cpp/container/vector/reserve)

include/sparrow/buffer.hpp
JohanMabille: Because of the min macro defined on Windows, but we can remove it if we define `NOMINMAX` and propagate it.

include/sparrow/buffer.hpp
```diff
@@ -18,454 +18,870 @@
...
+        m_data.p_storage_end = m_data.p_begin + n;
+    }
 
-        buffer_view(buffer_view&&) = default;
-        buffer_view& operator=(buffer_view&&) = default;
-    };
+    template <class T>
+    constexpr void buffer_base<T>::assign_storage(pointer p, size_type n, size_type cap)
+    {
+        m_data.p_begin = p;
```
jjerphan, line 379:
```suggestion
        assert(n <= cap);
        m_data.p_begin = p;
```

include/sparrow/buffer.hpp
```diff
@@ -18,454 +18,870 @@
...
+            constexpr buffer_data& operator=(buffer_data&&) noexcept;
+        };
+
+        buffer_base() = default;
+        
+        template <allocator A>
+        constexpr buffer_base(const A& a) noexcept;
+
+        template <allocator A = allocator_type>
+        buffer_base(size_type n, const A& a = A());
```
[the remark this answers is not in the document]
jjerphan: Are there any reasons for it not to be `constexpr`?

include/sparrow/buffer.hpp
jjerphan, line 1: This file has to be clang-formatted.

include/sparrow/buffer.hpp
```diff
@@ -18,454 +18,870 @@
...
     }
 
     template <class T>
-    void buffer<T>::resize(size_type n)
+    template <class It>
+    constexpr auto buffer<T>::allocate_and_copy(size_type n, It first, It last) -> pointer
     {
-        // TODO: add capacity, resize if growing only and define a shrink_to_fit method
-        if (n != base_type::size())
+        pointer p = this->allocate(n);
```
jjerphan, line 808: Must it be present in the exception handling bellow?

include/sparrow/buffer.hpp
```diff
@@ -18,454 +18,870 @@
...
+        try
         {
-            buffer<T> tmp(n);
-            const size_type copy_size = std::min(base_type::size(), n);
-            std::copy(base_type::data(), base_type::data() + copy_size, tmp.data());
-            base_type::swap(tmp);
+            copy_initialize(first, last, p, get_allocator());
         }
+        catch(...)
```
jjerphan, line 813: If l.808 is exception-handled, this this be more specific?

jjerphan (commented):
LGTM modulo a few last comments.

include/sparrow/buffer_view.hpp
```diff
@@ -0,0 +1,270 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include "sparrow/buffer.hpp"
+
+namespace sparrow
+{
+    /*
+     * @class buffer_view
```
Klaim, line 22: You can remove this and rely on the autobrief syntax?

include/sparrow/buffer_view.hpp
```diff
@@ -0,0 +1,270 @@
...
+        using size_type = std::size_t;
+        using difference_type = std::ptrdiff_t;
+
+        using iterator = pointer_iterator<pointer>;
+        using const_iterator = pointer_iterator<const_pointer>;
+        using reverse_iterator = std::reverse_iterator<iterator>;
+        using const_reverse_iterator = std::reverse_iterator<const_iterator>;
+
+        explicit buffer_view(buffer<T>& buffer);
+        buffer_view(pointer p, size_type n);
```
Klaim, line 45: To be complete we also need a `subrange(...)` function and a constructor that takes a pair of iterators (they should be template parameters constrained by concept, not `buffer_view::iterator`).
These are often used with view types (see `std::span`, which also provides `first` and `last` subview functions).

include/sparrow/buffer_view.hpp
```diff
@@ -0,0 +1,270 @@
...
+        reverse_iterator rbegin();
+        reverse_iterator rend();
+
+        const_reverse_iterator rbegin() const;
+        const_reverse_iterator rend() const;
+        const_reverse_iterator crbegin() const;
+        const_reverse_iterator crend() const;
+
+        void swap(buffer_view& rhs) noexcept;
+        bool equal(const buffer_view& rhs) const;
```
Klaim, line 82: This is not standard but is public, could you document what it does exactly? It compares elements? Is it necessary when we have `std::equal` and `std::ranges::equal` already? Or did you intend it to be private?

include/sparrow/buffer_view.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+    private:
+
+        pointer p_data = nullptr;
+        size_type m_size = 0u;
+    };
+
+    template <class T>
+    bool operator==(const buffer_view<T>& lhs, const buffer_view<T>& rhs);
```
Klaim, line 91: Why not declare it in `buffer_view` as friend?

BTW should we enable comparison with another type of (contiguous?) range?

include/sparrow/buffer_view.hpp
```diff
@@ -0,0 +1,270 @@
...
+#include "sparrow/buffer.hpp"
+
+namespace sparrow
+{
+    /*
+     * @class buffer_view
+     * @brief Object that references but does not own a piece of contiguous memory
+     */
+    template <class T>
+    class buffer_view
```
Klaim, line 26: How is this type different from `std::span`?

test/test_buffer.cpp
```diff
@@ -161,32 +168,86 @@ namespace sparrow
             CHECK_EQ(b3.data()[idx], expected_value);
         }
 
-        TEST_CASE("equality comparison")
+        // Iterators
+
```
Klaim, line 172: Missing tests: static check that
- for a given set of T, buffer_view<T> is a contiguous range
- buffer_view<T>::iterator is the appropriate type of iterator

test/test_buffer.cpp
```diff
@@ -161,32 +168,86 @@ namespace sparrow
...
+            const std::size_t size = 8u;
+            buffer_test_type b(make_test_buffer(size), size);
+            auto iter = b.rbegin();
+            auto citer = b.crbegin();
+            for (std::size_t i = b.size(); i != 0u; --i)
+            {
+                CHECK_EQ(*iter++, b[i - 1]);
+                CHECK_EQ(*citer++, b[i - 1]);
+            }
+            CHECK_EQ(iter, b.rend());
```
Klaim, line 199: these should also match some iterators obtained through `begin/end`

test/test_buffer.cpp
Klaim: Actually, I dont see any test specific to buffer_view?

include/sparrow/dynamic_bitset.hpp
```diff
@@ -20,6 +20,7 @@
 #include <ranges>
 
 #include "sparrow/buffer.hpp"
+#include "sparrow/buffer_view.hpp"
```
Klaim, line 23: I dont see it used here?

include/sparrow/buffer_view.hpp
Klaim: `operator!=` is missing?

include/sparrow/buffer.hpp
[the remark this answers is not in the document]
JohanMabille: Nope, I thought I fixed it Oo

include/sparrow/buffer_view.hpp
JohanMabille: This was to avoid declaring `operator==` as friend, but I agree this is not standard and should be removed.

include/sparrow/buffer.hpp
JohanMabille: No, the exception handling is to make sure the memory is deallocated if the initialization throws. If the allcoation throws, we have nothing to deallocate and we let the user decide what to do.

include/sparrow/buffer_view.hpp
JohanMabille: `operator!=` is now (C++20) automatically generated when you provide `operator==`

jjerphan: > `operator!=` is missing?

In C++20, it is automatically defined after `operator==`'s definition.

include/sparrow/buffer_view.hpp
JohanMabille: Not sure, I will set up a skeleton for the doc and check whether we can do it in a dedicated PR. If so, I will remove all of them.

include/sparrow/buffer_view.hpp
JohanMabille: Maybe we can do that in a dedicated PR? This piece of code was just moved from the initial `buffer.hpp` file. I'll add a todo for that.

include/sparrow/buffer_view.hpp
JohanMabille: They are very similar indeed. However, `std::span` is missing some APIs in C++20 (like `cbegin` / `cend` ), and we may want to provide additional std;:vector like APIs.

test/test_buffer.cpp
JohanMabille: These are the tests of the `buffer` class; the test suite of `buffer_view` now starts at line 315 and remains unchanged.

include/sparrow/dynamic_bitset.hpp
JohanMabille: In the `dynamic_bitset_view` class.

jjerphan (approved):
LGTM up to file formatting.

include/sparrow/buffer_view.hpp
Klaim: I feel that this and the iterator pair constructor (which should be what the subrange is built with) not being there makes it incomplete. But ok, it can be added in another PR although please do it immediately after this one then?

include/sparrow/buffer_view.hpp
Klaim: Could you add that clarification in the comments, so that we keep a trace in the code?

Klaim (approved):
Other than the details I pointed LGTM

PR #77, opened 2024-04-17
Contracts header with assertion macros specific to Sparrow

Klaim:
This provides macros (extracted from my library NoContracts) for contract assertions checks, named with SPARROW_ so that control over enabling these assertions is let to the user.

The checks are enabled by default and all C `assert(...)` calls have been replaced.

### DIFF
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -220,28 +220,28 @@ namespace sparrow
     template <class T>
     auto buffer_base<T>::front() -> reference
     {
-        SPARROW_ASSERT_TRUE(!empty());
+        SPARROW_ASSERT_FALSE(empty());
         return data()[0];
     }
 
     template <class T>
     auto buffer_base<T>::front() const -> const_reference
     {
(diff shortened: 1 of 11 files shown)


JohanMabille (commented):
Same remarl as Alexis, I think `SPARROW_ASSERT_TRUE(!expr)` should be replaced with `SPARROW_ASSERT_FALSE`.

Klaim:
OK I made a big replace operation to do that so now I'll review each assert see if the other option isnt better 👍🏽

Alex-PLACET (approved):
If you have the reference, I can say:
"Je contracte!"

include/sparrow/buffer.hpp
jjerphan, line 1: Just a last comment: this file needs formatting and a few other ones as well it seems.

JohanMabille (approved):

include/sparrow/buffer.hpp
JohanMabille: The current formatting rules mess the macro definitions, making them very hard to read. We should update the rules in a dedicated PR and then format.

PR #81, opened 2024-04-22
Support clang++ from conda-forge

jjerphan:
Even-though conda-forge's toolchain allows for using clang++ 18, libc++ 16 is the latest version available (see https://github.com/conda-forge/libcxx-feedstock/pull/131).

The `std::lexicographical_compare_three_way` has only been implemented in libc++ 17, so we use the vendored implementation when libc++ < 17 is used.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -13,7 +13,7 @@ defaults:
 jobs:
   build:
     runs-on: macos-${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


JohanMabille (commented):
We discussed dropping support for Apple Clang, is it still relevant? If so, we should probably remove the corresponding macro; otherwise, we should have an additional parameter in the matrix build to test bothe Apple Clang and clang 17.

JohanMabille (approved):

PR #82, opened 2024-04-23
ci: Test GCC 12 on Linux

jjerphan:
GCC / stdlibc++ 13 is relatively recent for most toolchains.

Let's also support GCC / stdlibc++ 12.


Klaim (approved):
LGTM

JohanMabille (approved):

Issue #34, opened 2024-03-21
Implement insert, erase and push_back methods in buffer


JohanMabille:
This should be done after #45 is complete.

PR #86, opened 2024-05-06
Add insert, emplace, erase, push_back and pop_back

Alex-PLACET:
Implement #34


include/sparrow/buffer.hpp
```diff
@@ -686,6 +702,132 @@ namespace sparrow
...
+    {
+        SPARROW_ASSERT_TRUE(cbegin() <= pos && pos <= cend())
+        const size_type num_elements = last - first;
+        const size_type new_size = size() + num_elements;
+        const size_type offset = pos - cbegin();
+        reserve(new_size);
+        auto& data = get_data();
+        const pointer p = data.p_begin + offset;
+        std::move_backward(p, data.p_end, data.p_end + num_elements);
+        std::copy(first, last, p);
```
Klaim, line 754: Potential future optimization: a copy_or_move function which if the conditions are right (range moved in) will move instead of copy, otherwise copy. There was an implemetnation in Boost.Move but not sure if `std::copy` doesnt do that already under specific conditions.

include/sparrow/buffer.hpp
```diff
@@ -686,6 +702,132 @@ namespace sparrow
...
+        const size_type sz = size();
+        const size_type off = pos - cbegin();
+        if (sz <= capacity())
+        {
+            reserve(sz + 1);
+        }
+        pointer p = get_data().p_begin + off;
+        if (p != get_data().p_end)
+        {
+            alloc_traits::construct(get_allocator(), get_data().p_end, std::move(*(get_data().p_end - 1)));
```
Klaim, line 780: I'm confused, why are both construct and move_backward on the same element necessary here?

test/test_buffer.cpp
```diff
@@ -26,9 +26,11 @@ namespace sparrow
     using buffer_test_type = buffer<uint8_t>;
     using view_test_type = buffer_view<uint8_t>;
 
+    using non_trivial_buffer_test_type = buffer<std::string>;
+
     namespace
     {
-        uint8_t* make_test_buffer(std::size_t size, uint8_t start_value = 0)
+        auto make_test_buffer(std::size_t size, uint8_t start_value = 0) -> uint8_t*
```
Klaim, line 33: I'm not completely sure but I suspect all the usage of `uint8_t` here (including previous code) can be replaced by `sparrow::byte_t` ?

test/test_buffer.cpp
```diff
@@ -350,6 +352,156 @@ namespace sparrow
             buffer_test_type b3(make_test_buffer(size2), size2);
             CHECK(b1 != b3);
         }
+
+        TEST_CASE("emplace")
```
Klaim, line 356: All the insertion operations needs to be tested with these conditions:
- when the container is empty (just constructed)
- when it has values
  - insert at the beginning (the only one tested here for example)
  - insert in the middle
  - insert at the end

test/test_buffer.cpp
```diff
@@ -350,6 +352,156 @@ namespace sparrow
...
+                CHECK_EQ(b[1], 102);
+                CHECK_EQ(b[2], 103);
+                CHECK_EQ(b[3], 0);
+                CHECK_EQ(b[4], 1);
+                CHECK_EQ(b[5], 2);
+                CHECK_EQ(b[6], 3);
+            }
+        }
+
+        TEST_CASE("erase")
```
Klaim, line 455: Erasure also needs to be tested at begin, end and in the middle of container of either 1 or n > 1 elements.

test/test_buffer.cpp
```diff
@@ -350,6 +352,156 @@ namespace sparrow
...
+        {
+            SUBCASE("The new element is initialized as a copy of value")
+            {
+                constexpr std::size_t size = 4u;
+                buffer_test_type b(make_test_buffer(size), size);
+
+                constexpr uint8_t expected_value = 101;
+                b.push_back(expected_value);
+                REQUIRE_EQ(b.size(), size + 1);
+                CHECK_EQ(b[size], expected_value);
```
Klaim, line 491: Add a check that `end()` points to the element after the added element. Same for all `push_back` tests.

test/test_buffer.cpp
```diff
@@ -350,6 +352,156 @@ namespace sparrow
...
+            }
+
+            SUBCASE("Value is moved into the new element.")
+            {
+                constexpr std::size_t size = 4u;
+                buffer_test_type b(make_test_buffer(size), size);
+
+                uint8_t expected_value = 101;
+                b.push_back(std::move(expected_value));
```
Klaim, line 500: This will effectively copy. Try with a move-only type instead, just to detect issues related to move.

include/sparrow/buffer.hpp
Alex-PLACET: The two operations are doing different things:

1. `alloc_traits::construct(get_allocator(), get_data().p_end, std::move(*(get_data().p_end - 1)));` is constructing a new object at the location `get_data().p_end` by moving the object that is currently at `get_data().p_end - 1`. After this operation, the object at `get_data().p_end - 1` is in a valid but unspecified state.

2. `std::move_backward(p, get_data().p_end - 1, get_data().p_end);` is moving the objects in the range `[p, get_data().p_end - 1)` to the range `[p + 1, get_data().p_end)`. This effectively shifts the objects one position to the right.

The reason both are necessary is because the first operation creates space for the second operation to move the objects into. Without the first operation, the second operation would overwrite the last object in the range.

test/test_buffer.cpp
Alex-PLACET: I'm ok to change that.
Are you ok @JohanMabille ?

Alex-PLACET: I finally changed that type to use a int32_t to avoid a sweet spot and to catch more errors in tests.

include/sparrow/buffer.hpp
```diff
@@ -686,6 +705,160 @@ namespace sparrow
...
+
+    template <class T>
+    template <class... Args>
+    constexpr buffer<T>::iterator buffer<T>::emplace(const_iterator pos, Args&&... args)
+    {
+        SPARROW_ASSERT_TRUE(cbegin() <= pos)
+        SPARROW_ASSERT_TRUE(pos <= cend())
+        const size_type sz = size();
+        const size_type off = pos - cbegin();
+        if (sz <= capacity())
```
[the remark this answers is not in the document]
Alex-PLACET: You right it's useless. I removed the statement

JohanMabille (approved):

include/sparrow/buffer.hpp
```diff
@@ -686,6 +702,132 @@ namespace sparrow
...
+        }
+        return begin() + __offset;
+    }
+
+    template <class T>
+    template <class InputIt>
+    constexpr auto buffer<T>::insert(const_iterator pos, InputIt first, InputIt last) -> iterator
+    {
+        SPARROW_ASSERT_TRUE(cbegin() <= pos && pos <= cend())
+        const size_type num_elements = last - first;
```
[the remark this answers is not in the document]
Klaim: I see a lot of code still not using `std::distance` when I suspect it would be best. Or do you have reasons to not use it?

test/test_buffer.cpp
Klaim: I dont understand why specifically `int32_t` and not the type that represent bytes in the library? (also note that `start_value` needs to have that type too?)

Alex-PLACET: I choose int32_t because it's 4 bytes and not a single byte as previously and I expect to catch more issues with this case.
About `start_value`, it's the case, code is outdated in the preview.

include/sparrow/buffer.hpp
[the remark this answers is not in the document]
Alex-PLACET: Fixed btw

include/sparrow/buffer.hpp
```diff
@@ -686,6 +705,156 @@ namespace sparrow
...
+        return emplace(pos, std::move(value));
+    }
+
+    template <class T>
+    constexpr auto buffer<T>::insert(const_iterator pos, size_type count, const T& value) -> iterator
+    {
+        SPARROW_ASSERT_TRUE(cbegin() <= pos);
+        SPARROW_ASSERT_TRUE(pos <= cend());
+
+        const difference_type __offset = std::distance(cbegin(), pos);
```
Klaim, line 730: Double underscore names are reserved by the standard, maybe just change to `offset`?

test/test_buffer.cpp
```diff
@@ -23,17 +23,34 @@
...
             return res;
         }
+
+        auto make_test_buffer_non_trivial(std::size_t size) -> std::string*
+        {
+            std::vector<std::string> vec(size);
+            vec.insert(vec.begin(), vec.end(), vec.end());
+            for (std::size_t i = 0; i < size; ++i)
+            {
+                vec[i] = std::to_string(i);
```
Klaim, line 46: Not sure it's important but just in case: `to_string` is subject to locale and that also makes it inefficient, which shouldnt impact things here, but just so you know there is `to_chars` series of algorithms that solve both issues. Probably not worth here as this is all tests.

test/test_buffer.cpp
```diff
@@ -138,7 +155,7 @@ namespace sparrow
         TEST_CASE("front")
         {
             const std::size_t size = 4u;
-            const std::uint8_t expected_value = 3u;
+            const std::int32_t expected_value = 3u;
```
Klaim, line 158: Note that the `u` is not necessary anymore here (same with other values below)

Klaim (approved):
Other than the necessary renaming, LGTM!

test/test_buffer.cpp
Alex-PLACET: Yes, usualy I use Boost to do that but as you said, it's just the unit tests and with std::_to_string the code is short and expressive.

test/test_buffer.cpp
Alex-PLACET: Ok done

PR #90, opened 2024-05-14
Add warnings options

Alex-PLACET:
Implement #88

### DIFF
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -338,7 +338,7 @@ namespace sparrow
     template <class L>
     auto const_reference_proxy<L>::value() const -> const_reference
     {
-        SPARROW_ASSERT_TRUE(has_value());
+        SPARROW_ASSERT_TRUE(has_value())
         return m_val_ref;
     }
 
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -18,11 +18,12 @@
 #include <optional>
 #include <vector>
 
-#include "sparrow/contracts.hpp"
 #include "sparrow/buffer.hpp"
+#include "sparrow/contracts.hpp"
 #include "sparrow/data_type.hpp"
 #include "sparrow/dynamic_bitset.hpp"
 #include "sparrow/memory.hpp"
+
 
 namespace sparrow
 {
@@ -367,14 +368,14 @@ namespace sparrow
     template <class L>
     auto reference_proxy<L>::value() -> value_type&
     {
-        SPARROW_ASSERT_TRUE(has_value());
+        SPARROW_ASSERT_TRUE(has_value())
         return m_val_ref;
     }
 
     template <class L>
     auto reference_proxy<L>::value() const -> const value_type&
     {
-        SPARROW_ASSERT_TRUE(has_value());
+        SPARROW_ASSERT_TRUE(has_value())
         return m_val_ref;
     }
 
--- a/include/sparrow/array_data.hpp
+++ b/include/sparrow/array_data.hpp
@@ -338,7 +338,7 @@ namespace sparrow
     template <class L>
     auto const_reference_proxy<L>::value() const -> const_reference
     {
-        SPARROW_ASSERT_TRUE(has_value())
+        SPARROW_ASSERT_TRUE(has_value());
         return m_val_ref;
     }
 
@@ -368,14 +368,14 @@ namespace sparrow
     template <class L>
     auto reference_proxy<L>::value() -> value_type&
     {
-        SPARROW_ASSERT_TRUE(has_value())
+        SPARROW_ASSERT_TRUE(has_value());
         return m_val_ref;
     }
 
     template <class L>
     auto reference_proxy<L>::value() const -> const value_type&
     {
-        SPARROW_ASSERT_TRUE(has_value())
+        SPARROW_ASSERT_TRUE(has_value());
         return m_val_ref;
     }
 
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -20,10 +20,11 @@
 #include <iterator>
 #include <stdexcept>
 
-#include "sparrow/contracts.hpp"
 #include "sparrow/allocator.hpp"
+#include "sparrow/contracts.hpp"
 #include "sparrow/iterator.hpp"
 #include "sparrow/mp_utils.hpp"
+
 
 namespace sparrow
 {
@@ -666,7 +667,10 @@ namespace sparrow
                 std::make_move_iterator(get_data().p_end)
             );
             destroy(get_data().p_begin, get_data().p_end, get_allocator());
-            this->deallocate(get_data().p_begin, static_cast<size_type>(get_data().p_storage_end - get_data().p_begin));
+            this->deallocate(
+                get_data().p_begin,
+                static_cast<size_type>(get_data().p_storage_end - get_data().p_begin)
+            );
             this->assign_storage(tmp, old_size, new_cap);
         }
     }
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -501,42 +501,42 @@ namespace sparrow
     template <class T>
     constexpr auto buffer<T>::operator[](size_type i) -> reference
     {
-        SPARROW_ASSERT_TRUE(i < size());
+        SPARROW_ASSERT_TRUE(i < size())
         return get_data().p_begin[i];
     }
 
     template <class T>
     constexpr auto buffer<T>::operator[](size_type i) const -> const_reference
     {
-        SPARROW_ASSERT_TRUE(i < size());
+        SPARROW_ASSERT_TRUE(i < size())
         return get_data().p_begin[i];
     }
 
     template <class T>
     constexpr auto buffer<T>::front() -> reference
     {
-        SPARROW_ASSERT_FALSE(empty());
+        SPARROW_ASSERT_FALSE(empty())
         return *(get_data().p_begin);
     }
 
     template <class T>
     constexpr auto buffer<T>::front() const -> const_reference
     {
-        SPARROW_ASSERT_FALSE(empty());
+        SPARROW_ASSERT_FALSE(empty())
         return *(get_data().p_begin);
     }
 
     template <class T>
     constexpr auto buffer<T>::back() -> reference
     {
-        SPARROW_ASSERT_FALSE(empty());
+        SPARROW_ASSERT_FALSE(empty())
         return *(get_data().p_end - 1);
     }
 
     template <class T>
     constexpr auto buffer<T>::back() const -> const_reference
     {
-        SPARROW_ASSERT_FALSE(empty());
+        SPARROW_ASSERT_FALSE(empty())
         return *(get_data().p_end - 1);
     }
 
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -371,7 +371,7 @@ namespace sparrow
     template <class T>
     constexpr void buffer_base<T>::assign_storage(pointer p, size_type n, size_type cap)
     {
-        SPARROW_ASSERT_TRUE(n <= cap)
+        SPARROW_ASSERT_TRUE(n <= cap);
         m_data.p_begin = p;
         m_data.p_end = p + n;
         m_data.p_storage_end = p + cap;
@@ -502,42 +502,42 @@ namespace sparrow
     template <class T>
     constexpr auto buffer<T>::operator[](size_type i) -> reference
     {
-        SPARROW_ASSERT_TRUE(i < size())
+        SPARROW_ASSERT_TRUE(i < size());
         return get_data().p_begin[i];
     }
 
     template <class T>
     constexpr auto buffer<T>::operator[](size_type i) const -> const_reference
     {
-        SPARROW_ASSERT_TRUE(i < size())
+        SPARROW_ASSERT_TRUE(i < size());
         return get_data().p_begin[i];
     }
 
     template <class T>
     constexpr auto buffer<T>::front() -> reference
     {
-        SPARROW_ASSERT_FALSE(empty())
+        SPARROW_ASSERT_FALSE(empty());
         return *(get_data().p_begin);
     }
 
     template <class T>
     constexpr auto buffer<T>::front() const -> const_reference
     {
-        SPARROW_ASSERT_FALSE(empty())
+        SPARROW_ASSERT_FALSE(empty());
         return *(get_data().p_begin);
     }
 
     template <class T>
     constexpr auto buffer<T>::back() -> reference
     {
-        SPARROW_ASSERT_FALSE(empty())
+        SPARROW_ASSERT_FALSE(empty());
         return *(get_data().p_end - 1);
     }
 
     template <class T>
     constexpr auto buffer<T>::back() const -> const_reference
     {
-        SPARROW_ASSERT_FALSE(empty())
+        SPARROW_ASSERT_FALSE(empty());
         return *(get_data().p_end - 1);
     }
 
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -312,7 +312,7 @@ namespace sparrow
     template <class T>
     buffer_base<T>::~buffer_base()
     {
-        deallocate(m_data.p_begin, (m_data.p_storage_end - m_data.p_begin));
+        deallocate(m_data.p_begin, static_cast<size_type>(m_data.p_storage_end - m_data.p_begin));
     }
 
     template <class T>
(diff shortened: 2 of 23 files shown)


include/sparrow/dynamic_bitset.hpp
```diff
@@ -561,7 +577,11 @@ namespace sparrow
 
     template <std::integral T>
     dynamic_bitset<T>::dynamic_bitset(size_type n, value_type value)
-        : base_type(storage_type(this->compute_block_count(n), value ? ~block_type(0) : 0), n, value ? 0u : n)
+        : base_type(
+              storage_type(this->compute_block_count(n), value ? block_type(~block_type(0)) : 0),
```
jjerphan, line 581: For consistency?
```suggestion
              storage_type(this->compute_block_count(n), value ? block_type(~block_type(0)) : block_type(0)),
```

test/test_iterator.cpp
```diff
@@ -263,6 +263,11 @@ namespace sparrow
 
             const_iterator iter{&(a[0])};
             CHECK_EQ(*iter, a[0]);
+
+            iterator iter2{&(a[0])};
+            CHECK_EQ(*iter2, a[0]);
+            *iter2 = 3;
+            CHECK_EQ(a[0], 3);
```
jjerphan, line 270: This change looks unrelated to the scope of this PR, but it being part of it is OK to me.

test/test_typed_array.cpp
```diff
@@ -180,8 +180,9 @@ TEST_SUITE("typed_array")
             const typed_array<T> ta{array_data};
             const auto bitmap = ta.bitmap();
             REQUIRE_EQ(bitmap.size(), n - offset);
-            for (size_t i = 0; i < bitmap.size() - 1; ++i)
+            for (int32_t i = 0; i < static_cast<int32_t>(bitmap.size()) -1; ++i)
```
jjerphan, line 183: What is the reason for using `int32_t`, here and bellow?

test/CMakeLists.txt
```diff
@@ -54,6 +54,69 @@ add_executable(${test_target} ${SPARROW_TESTS_SOURCES})
...
+            -Wsign-conversion
+            -Wmisleading-indentation
+            -Wduplicated-cond
+            -Wduplicated-branches
+            -Wlogical-op
+            -Wnull-dereference
+            -Wuseless-cast
+            -Wdouble-promotion
+            -Wformat=2 
+            -Wimplicit-fallthrough
```
jjerphan, line 92: Naive question: do you think it would be worth factorizing flags used by  both `gcc` and `clang`?

include/sparrow/buffer.hpp
jjerphan, line 1: It seems this file needs reformatting.

jjerphan (commented):
LGTM modulo a few comments and questions.

include/sparrow/dynamic_bitset.hpp
Alex-PLACET: Fix a warning about a conversion

test/CMakeLists.txt
jjerphan: How about defining a list for the common ones?

Not that this suggestion need not be addressed at all.

test/test_iterator.cpp
Alex-PLACET: You right, this tests was missing

include/sparrow/dynamic_bitset.hpp
jjerphan: My remark was unclear: based [on this change on l.525](https://github.com/xtensor-stack/sparrow/pull/90/files#diff-eb6ccf28ffe0d12cb1cfb2b3d4c3871ee6430fd1d558b86880a69e62788f7f33R541), I was wondering if this line also should be modified similarly?

test/test_typed_array.cpp
Alex-PLACET: bitmap[] argument is a range_difference_t (aka 'long long'), and so the conversion changes the signedness

jjerphan (approved):
LGTM.

It seems that a few files aren't formatted (this need not be addressed IMO).

include/sparrow/mp_utils.hpp
```diff
@@ -192,7 +192,7 @@ namespace sparrow::mpl
     ///          or if the list is empty; `false` otherwise.
     template <class Predicate, template <class...> class L, class... T>
         requires any_typelist<L<T...>> and (callable_type_predicate<Predicate, T> && ...)
-    consteval bool all_of(L<T...> list, Predicate predicate)
+    consteval bool all_of([[maybe_unused]]L<T...> list, [[maybe_unused]] Predicate predicate)
```
Klaim, line 195: Note that it might be simpler here to just remove the name of `list`.
Also note that `predicate` will always be refered to, so it doesnt  need to be `maybe_unused`.

test/CMakeLists.txt
```diff
@@ -54,6 +54,44 @@ add_executable(${test_target} ${SPARROW_TESTS_SOURCES})
 target_link_libraries(${test_target} PRIVATE sparrow doctest::doctest)
 add_test(NAME ${test_target} COMMAND ${test_target})
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
Klaim, line 57: I would like that we try all the flags recommended by Jason Turner: https://github.com/cpp-best-practices/cppbestpractices/blob/master/02-Use_the_Tools_Available.md#compilers
and then progressively remove what we dont want.

For MSVC we also need to add `/Zc:__cplusplus` to make sure the `__cplusplus` define has the right value, see: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

Finally, shouldnt we have these flags in cmake toolchains instead of here? Doesnt this will affect users? (not sure what's the best practice regarding this)

include/sparrow/contracts.hpp
```diff
@@ -62,22 +62,44 @@
...
 # elif defined(SPARROW_CONTRACTS_USE_STD_FORMAT) && SPARROW_CONTRACTS_USE_STD_FORMAT == 1
 #   include <format>
 #   include <cstdio>
 #   define SPARROW_CONTRACTS_LOG_FAILURE( expr__, message__ ) \
-      ::fprintf(stderr, ::std::format("Assertion Failed ({}:{}): {} - ({} is wrong)\n", __FILE__, __LINE__, message__, #expr__ ).c_str())
+      do { \
+            SPARROW_CONTRACTS_IGNORE_WARNINGS; \
+            ::fprintf(stderr, ::std::format("Assertion Failed ({}:{}): {} - ({} is wrong)\n", __FILE__, __LINE__, message__, #expr__ ).c_str()); \
+            SPARROW_CONTRACTS_RESTORE_WARNINGS; \
+      } while (0);
```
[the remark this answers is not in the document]
jjerphan: > BTW why the need for do/while? Why not just write a block?

An explanation is given in [this thread](https://groups.google.com/g/comp.lang.c/c/xGZxls194mI/m/dEIpTKz2okMJ).

include/sparrow/dynamic_bitset.hpp
```diff
@@ -740,38 +760,38 @@ namespace sparrow
...
+        SPARROW_ASSERT_TRUE(m_index < bitset_type::s_bits_per_block)
     }
 
     template <class B, bool is_const>
     void bitset_iterator<B, is_const>::advance(difference_type n)
     {
         if (n >= 0)
         {
-            if (m_index + n < bitset_type::s_bits_per_block)
+            if (std::cmp_less(n, bitset_type::s_bits_per_block - m_index))
```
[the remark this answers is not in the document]
Alex-PLACET: The issue was the comparison between a signed and unsigned values.

include/sparrow/dynamic_bitset.hpp
```diff
@@ -522,7 +538,7 @@ namespace sparrow
     {
         const size_type old_block_count = m_buffer.size();
         const size_type new_block_count = compute_block_count(n);
-        const block_type value = b ? ~block_type(0) : block_type(0);
+        const block_type value = b ? block_type(~block_type(0)) : block_type(0);
```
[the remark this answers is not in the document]
Alex-PLACET: > error: operand of ? changes signedness: 'int' to 'const block_type' (aka 'const unsigned char')

test/CMakeLists.txt
Alex-PLACET: I activated all warnings, just to test, I finaly deactivated a dozen of them. It's maybe easier to maintain than activating some of them and forgot some.
I let you check the code (I commented the flags to explicitly activate the warnings)

include/sparrow/data_type.hpp
```diff
@@ -25,7 +25,24 @@
 // TODO: use exclusively `std::float16_t etc. once we switch to c++23, see
 // https://en.cppreference.com/w/cpp/types/floating-point
 #if __cplusplus <= 202002L
+#if defined(__GNUC__)
```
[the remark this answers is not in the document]
Alex-PLACET: I forgot to activate the warning as error for msvc, it's fixed now.
I deactivate theses warnings as the header is a 3rd party and we don't have any unit test to validate it, so I prefer to not change the code.

include/sparrow/contracts.hpp
```diff
@@ -62,22 +62,44 @@
...
+#define SPARROW_CONTRACTS_RESTORE_WARNINGS
+#endif
+
 #ifndef SPARROW_CONTRACTS_LOG_FAILURE
 # if defined(SPARROW_CONTRACTS_USE_STD_PRINT) && SPARROW_CONTRACTS_USE_STD_PRINT == 1
 #   include <print>
 #   include <cstdio>
 #   define SPARROW_CONTRACTS_LOG_FAILURE( expr__, message__ ) \
-      ::std::print(stderr, "Assertion Failed ({}:{}): {} - ({} is wrong)\n", __FILE__, __LINE__, message__, #expr__ )
+      ::std::print(stderr, "Assertion Failed ({}:{}): {} - ({} is wrong)\n", __FILE__, __LINE__, message__, #expr__ );
```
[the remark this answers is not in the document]
Alex-PLACET: Ok I reverted

include/sparrow/array_data.hpp
```diff
@@ -337,7 +338,7 @@ namespace sparrow
     template <class L>
     auto const_reference_proxy<L>::value() const -> const_reference
     {
-        SPARROW_ASSERT_TRUE(has_value());
+        SPARROW_ASSERT_TRUE(has_value())
```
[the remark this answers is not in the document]
Alex-PLACET: Ok, reverted 👍

include/sparrow/dynamic_bitset.hpp
```diff
@@ -344,7 +345,14 @@ namespace sparrow
         }
         else
         {
+#if defined(__GNUC__)
```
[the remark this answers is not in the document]
Alex-PLACET: Fixed 👍

include/sparrow/dynamic_bitset.hpp
```diff
@@ -500,7 +509,14 @@ namespace sparrow
         const size_type extra_bits = count_extra_bits();
         if (extra_bits != 0)
         {
+#if defined(__GNUC__)
```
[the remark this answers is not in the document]
Alex-PLACET: Fixed 👍

include/sparrow/mp_utils.hpp
Alex-PLACET: I fixed the issue for list by just removing the name, but on GCC, the predicate parameter is detected as not used

test/CMakeLists.txt
Alex-PLACET: About where to put these flags, I think we should keep them in our test as the user may not have the same requirements in terms of warnings as errors

include/sparrow/contracts.hpp
```diff
@@ -12,92 +12,149 @@
...
-#   include <cstdlib>
-#   include <cstdio>
-#   define SPARROW_CONTRACTS_LOG_FAILURE( expr__, message__ ) \
-      ::fprintf(stderr, "Assertion Failed (%s:%i): %s - (%s is wrong)\n", __FILE__, __LINE__, message__, #expr__ )
-# endif
+#    if defined(SPARROW_CONTRACTS_USE_STD_PRINT) && SPARROW_CONTRACTS_USE_STD_PRINT == 1
+#        include <cstdio>
+#        include <print>
+
+#        define SPARROW_CONTRACTS_LOG_FAILURE(expr__, message__)  \
```
Klaim, line 96: I think the formatting is a bit extreme for these function calls 😂 
Can you deactivate it for the 3 print calls and if you want to indent it in 2 lines it's fine? Right now it's header to read than the original.

include/sparrow/contracts.hpp
```diff
@@ -12,92 +12,149 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// NOTE: This is based upon v0.3.0 of NoContracts library https://github.com/Klaim/nocontracts/tree/0ffc183f7527213e3f0a57b8dae9befa7c0ca02c
-// Modifications: renamed macros
+// NOTE: This is based upon v0.3.0 of NoContracts library
+// https://github.com/Klaim/nocontracts/tree/0ffc183f7527213e3f0a57b8dae9befa7c0ca02c Modifications: renamed
```
Klaim, line 16:
```suggestion
// https://github.com/Klaim/nocontracts/tree/0ffc183f7527213e3f0a57b8dae9befa7c0ca02c
// Modifications: renamed macros
```

include/sparrow/contracts.hpp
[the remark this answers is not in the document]
Klaim: I see, not sure if it's still undefined behavior though. Maybe it's not because it's not an infinite loop. Ok lets try that.

include/sparrow/data_type.hpp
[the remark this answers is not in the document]
Klaim: Nice, could you add a comment about these warnings being specific to that 3rdpary header?

include/sparrow/mp_utils.hpp
Klaim: Looks like a bug? it's used in the return expression...
Maye add a "TODO: remove `[[mayb_unused]]` once gcc is fixed" in the comment you added there.

include/sparrow/contracts.hpp
Klaim: Maybe also deactivate the formatting around the `do/while` block 🥲

Alex-PLACET: Oups, sorry I didn't view this auto indent, I will add a NOLINT block

include/sparrow/contracts.hpp
Alex-PLACET: Done

include/sparrow/data_type.hpp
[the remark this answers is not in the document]
Alex-PLACET: Done

test/CMakeLists.txt
```diff
@@ -54,6 +54,74 @@ add_executable(${test_target} ${SPARROW_TESTS_SOURCES})
...
+        # /w14555 # expression has no effect; expected expression with side- effect
+        # /w14619 # pragma warning: there is no warning number 'number'
+        # /w14640 # Enable warning on thread un-safe static member initialization
+        # /w14826 # Conversion from 'type1' to 'type_2' is sign-extended. This may cause unexpected runtime behavior.
+        # /w14905 # wide string literal cast to 'LPSTR'
+        # /w14906 # string literal cast to 'LPWSTR'
+        # /w14928 # illegal copy-initialization; more than one user-defined conversion has been implicitly applied
+        # /we4289 # nonstandard extension used: 'variable': loop control variable declared in the for-loop is used outside the for-loop scope
+        # /we4244 # conversion from 'type1' to 'type_2', possible loss of data
+        /Wall # all warnings
```
Klaim, line 82: This one is *not* recommended (see https://github.com/cpp-best-practices/cppbestpractices/blob/master/02-Use_the_Tools_Available.md#msvc ).
I suspect removing this will make the rest work better.

Klaim (approved):
LGTM except the msvc warning flags but it will be easier for me (who has a windows always available) to check these and patch after this is merge.

test/CMakeLists.txt
Klaim: No because `/Wall` activates warning from dependencies, which is what the Turner points. It's not something we can easilly turn off without dependency-specific flags. That's why it's not recommended with msvc. It's not like gcc/clang's `-Wall` which does *not* activate all warnings, but only recomended ones.

JohanMabille (approved):

test/CMakeLists.txt
Klaim: It is not enough to disable the dependencies included through `"` :/

PR #96, opened 2024-05-28
Replace /Wall on MSVC by explicit warnings, fix compilation error

Alex-PLACET:


include/sparrow/buffer.hpp
```diff
@@ -732,18 +731,13 @@ namespace sparrow
...
-            auto& data = get_data();
-            const auto new_ptr = data.p_begin + offset;
-            std::move_backward(new_ptr, data.p_end, data.p_end + count);
-            std::fill_n(new_ptr, count, value);
-            data.p_end += count;
+            reserve(size() + count);
+            const iterator it = std::next(begin(), offset);
+            std::move_backward(it, end(), std::next(end(), static_cast<difference_type>(count)));
+            std::fill_n(it, count, value);
+            get_data().p_end += count;
```
Alex-PLACET, line 740: Use of iterators instead of pointers to simply the code

test/test_buffer.cpp
```diff
@@ -15,6 +15,7 @@
 #include <iterator>
 #include <numeric>
 #include <ranges>
+#include <string>
```
Alex-PLACET, line 18: Fix because of the latest version of the MSVC

Issue #78, opened 2024-04-18
buffer_view missing method and constructor


JohanMabille:
The buffer_view is missing a `subrange` method and a constructor accepting a pair of iterators (see https://github.com/xtensor-stack/sparrow/pull/74#discussion_r1567099836)

Alex-PLACET:
What about just using https://en.cppreference.com/w/cpp/ranges/subrange ?

PR #97, opened 2024-05-31
Add buffer_view constructor with iterators and subrange methods

Alex-PLACET:
Implement #78

### DIFF
--- a/include/sparrow/buffer_view.hpp
+++ b/include/sparrow/buffer_view.hpp
@@ -48,10 +48,10 @@ namespace sparrow
         explicit buffer_view(buffer<T>& buffer);
         buffer_view(pointer p, size_type n);
 
-        template <class It>
-            requires std::input_iterator<It>
+        template <class It, class End>
+            requires std::contiguous_iterator<It> && std::sentinel_for<End, It>
                      && std::same_as<std::remove_const_t<std::iter_value_t<It>>, std::remove_const_t<T>>
-        buffer_view(It first, It last);
+        buffer_view(It first, End last);
 
         [[nodiscard]] bool empty() const noexcept;
         [[nodiscard]] size_type size() const noexcept;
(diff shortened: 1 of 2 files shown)


include/sparrow/buffer_view.hpp
```diff
@@ -47,14 +47,14 @@ namespace sparrow
...
-        // be template parameters constrained by concept, not buffer_view::iterator).
-        // These are often used with view types (see std::span, which also provides
-        // first and last subview functions).
 
-        bool empty() const noexcept;
-        size_type size() const noexcept;
+        template <class It>
+            requires std::input_iterator<It>
+                     && std::same_as<std::remove_const_t<std::iter_value_t<It>>, std::remove_const_t<T>>
+        buffer_view(It first, It last);
```
Klaim, line 54: I think it's preferable that `last` is allowed to be a different type than `It` because it could be a sentinel. Then to simplify this and the other functions with these requirements, probably a concept  `input_iterator_of<T>` or something like that could help?

Same remarks for the range constructor.

include/sparrow/buffer_view.hpp
```diff
@@ -116,6 +120,17 @@ namespace sparrow
     {
     }
 
+    template <class T>
+    template <class It>
+        requires std::input_iterator<It>
+                     && std::same_as<std::remove_const_t<std::iter_value_t<It>>, std::remove_const_t<T>>
+    buffer_view<T>::buffer_view(It first, It last)
+        : p_data(&*first)
```
Klaim, line 128: I dont think that's working if `It` is not a contiguous iterator?

BTW there is `addressof` and `pointer_to` that cold help with that expression but I'm not totally sure if it's necessary here.

include/sparrow/buffer_view.hpp
Klaim: As noted below, if `buffer_view` assumes it's necessarilly a `buffer`'s content being looked at then I suspect it should be a `contiguous_iterator` instead of `input_iterator` and that would fix my previous remark.

Alex-PLACET: You' right, `contiguous_iterator` is better in this case, I also added a requirement for the sentinel

include/sparrow/buffer_view.hpp
Alex-PLACET: Adressed too

Klaim (approved):
LGTM 👍🏽

Issue #46, opened 2024-03-27
Implement a buffer_adaptor class


JohanMabille:
Currently, the `array_data` stores `buffer<uint8_t>` objects for many reasons:
- it allows to store all the buffers in a `std::vector` without using any kind of type erasure
- it make array_data a simple class instead of a template class.
- array_data with children would have to be templated by the parameters of their children, making them very verbose and complicated.

However, when dealing with these buffers in the layout classes, we have to systematically cast the returned values (which is ver cumbersome), or restrict the usage of the class to the data method (which is very inconveninent).

A solution would be to implement a buffer_adaptor class, that would internally have a reference to a `buffer<T>`, and would provide the API of a `buffer<U>`.

PR #98, opened 2024-06-04
Add buffer_adaptor

Alex-PLACET:
Implement #46

### DIFF
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -20,6 +20,7 @@
 #include <memory>
 #include <stdexcept>
 #include <type_traits>
+#include <ranges>
 
 #include "sparrow/allocator.hpp"
 #include "sparrow/contracts.hpp"
@@ -148,6 +149,10 @@ namespace sparrow
 
         template <class It, allocator A = allocator_type>
         constexpr buffer(It first, It last, const A& a = A());
+
+        template <class Range, allocator A = allocator_type>
+        requires std::ranges::input_range<Range>
+        constexpr buffer(const Range& range, const A& a = A());
 
         ~buffer();
 
@@ -436,6 +441,15 @@ namespace sparrow
         : base_type(check_init_length(static_cast<size_type>(std::distance(first, last)), a), a)
     {
         get_data().p_end = copy_initialize(first, last, get_data().p_begin, get_allocator());
+    }
+
+    template <class T>
+    template <class Range, allocator A>
+    requires std::ranges::input_range<Range>
+    constexpr buffer<T>::buffer(const Range& range, const A& a)
+        : base_type(check_init_length(static_cast<size_type>(std::ranges::size(range)), a), a)
+    {
+        get_data().p_end = copy_initialize(std::ranges::begin(range), std::ranges::end(range), get_data().p_begin, get_allocator());
     }
 
     template <class T>
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -150,8 +150,7 @@ namespace sparrow
         template <class It, allocator A = allocator_type>
         constexpr buffer(It first, It last, const A& a = A());
 
-        template <class Range, allocator A = allocator_type>
-        requires std::ranges::input_range<Range>
+        template <std::ranges::input_range Range, allocator A = allocator_type>
         constexpr buffer(const Range& range, const A& a = A());
 
         ~buffer();
@@ -444,8 +443,7 @@ namespace sparrow
     }
 
     template <class T>
-    template <class Range, allocator A>
-    requires std::ranges::input_range<Range>
+    template <std::ranges::input_range Range, allocator A>
     constexpr buffer<T>::buffer(const Range& range, const A& a)
         : base_type(check_init_length(static_cast<size_type>(std::ranges::size(range)), a), a)
     {
--- a/include/sparrow/buffer.hpp
+++ b/include/sparrow/buffer.hpp
@@ -18,13 +18,14 @@
 #include <concepts>
 #include <iterator>
 #include <memory>
+#include <ranges>
 #include <stdexcept>
 #include <type_traits>
-#include <ranges>
 
 #include "sparrow/allocator.hpp"
 #include "sparrow/contracts.hpp"
 #include "sparrow/iterator.hpp"
+
 
 namespace sparrow
 {
@@ -447,7 +448,12 @@ namespace sparrow
     constexpr buffer<T>::buffer(const Range& range, const A& a)
         : base_type(check_init_length(static_cast<size_type>(std::ranges::size(range)), a), a)
     {
-        get_data().p_end = copy_initialize(std::ranges::begin(range), std::ranges::end(range), get_data().p_begin, get_allocator());
+        get_data().p_end = copy_initialize(
+            std::ranges::begin(range),
+            std::ranges::end(range),
+            get_data().p_begin,
+            get_allocator()
+        );
     }
 
     template <class T>
--- a/include/sparrow/buffer_adaptor.hpp
+++ b/include/sparrow/buffer_adaptor.hpp
@@ -130,7 +130,7 @@ namespace sparrow
         [[nodiscard]] buffer_reference::const_iterator get_buffer_reference_iterator(const_iterator pos);
 
         buffer_reference& m_buffer;
-        const size_type m_max_size;
+        size_type m_max_size;
     };
 
     template <typename To, typename From>
--- a/include/sparrow/buffer_adaptor.hpp
+++ b/include/sparrow/buffer_adaptor.hpp
@@ -1,0 +1,517 @@
+// Copyright 2024 Man Group Operations Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <cmath>
+#include <limits>
+#include <utility>
+
+#include "sparrow/buffer.hpp"
+#include "sparrow/contracts.hpp"
+
+namespace sparrow
+{
+    /**
+     * @brief Class which have internally a reference to a buffer<T> and provides the API of a buffer<U>
+     */
+    template <typename T, typename U>
+        requires(sizeof(U) < sizeof(T))
+    class buffer_adaptor
+    {
+    public:
+
+        using value_type = T;
+        using reference = value_type&;
+        using const_reference = const value_type&;
+        using pointer = value_type*;
+        using const_pointer = const value_type*;
+
+        using buffer_reference_value_type = U;
+        using buffer_reference = buffer<buffer_reference_value_type>;
+
+        using size_type = buffer_reference::size_type;
+        using difference_type = buffer_reference::difference_type;
+        using iterator = pointer_iterator<pointer>;
+        using const_iterator = pointer_iterator<const_pointer>;
+        using reverse_iterator = std::reverse_iterator<iterator>;
+        using const_reverse_iterator = std::reverse_iterator<const_iterator>;
+
+        buffer_adaptor(buffer_reference& buf)
+            : _buffer(buf)
+        {
+        }
+
+        constexpr pointer data() noexcept;
+        constexpr const_pointer data() const noexcept;
+
+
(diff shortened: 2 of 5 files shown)


include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+#include "sparrow/buffer.hpp"
+#include "sparrow/contracts.hpp"
+
+namespace sparrow
+{
+    /**
+     * @brief Class which have internally a reference to a buffer<T> and provides the API of a buffer<U>
+     */
+    template <typename T, typename U>
+        requires(sizeof(U) < sizeof(T))
```
JohanMabille, line 30: Why this requirement?

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+#include <limits>
+#include <utility>
+
+#include "sparrow/buffer.hpp"
+#include "sparrow/contracts.hpp"
+
+namespace sparrow
+{
+    /**
+     * @brief Class which have internally a reference to a buffer<T> and provides the API of a buffer<U>
```
JohanMabille, line 27: From the implementation and the requirements below, I guess you meant it holds a a reference to a `buffer<U>` and provides an API to a `buffer<T>`? what do you think of renaming the template parameters into From and To to avoid any confusion?

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+        constexpr void push_back(const value_type& value);
+
+        constexpr void pop_back();
+
+        constexpr void resize(size_type new_size);
+        constexpr void resize(size_type new_size, const value_type& value);
+
+    private:
+
+        static constexpr double _sizeof_ratio = static_cast<double>(sizeof(buffer_reference_value_type))
```
JohanMabille, line 126: We should avoid names starting with `_`, these are reserved for the standard.

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+                                                / static_cast<double>(sizeof(value_type));
+
+        static constexpr auto _type_size = static_cast<size_type>(1. / _sizeof_ratio);
```
JohanMabille, line 129: What do you think of the names `from_to_size_ratio` and `to_from_size_ratio`?
Also why having a `size_type` and not a double here?

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+
+        buffer_reference& _buffer;
+    };
+
+    template <typename T, typename U>
+        requires(sizeof(U) < sizeof(T))
+    constexpr typename buffer_adaptor<T, U>::buffer_reference::size_type
+    buffer_adaptor<T, U>::index_for_buffer(size_type idx) const noexcept
+    {
+        return static_cast<buffer_reference::size_type>(static_cast<double>(idx) / _sizeof_ratio);
```
JohanMabille, line 146: - I think you can use `size_type` here.
- I think `idx * to_from_size_ratio` would be more efficient than a division.

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+        static constexpr auto _type_size = static_cast<size_type>(1. / _sizeof_ratio);
+
+        static constexpr size_type _extra_buffervalue_type_size = sizeof(buffer_reference_value_type)
+                                                                  % sizeof(value_type);
+
+        [[nodiscard]] constexpr buffer_reference::size_type index_for_buffer(size_type idx) const noexcept;
+
+        [[nodiscard]] buffer_reference::const_iterator get_buffer_reference_iterator(const_iterator pos);
+
+        buffer_reference& _buffer;
```
JohanMabille, line 138: naming convention: => `m_buffer`

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+
+    template <typename T, typename U>
+        requires(sizeof(U) < sizeof(T))
+    constexpr buffer_adaptor<T, U>::size_type buffer_adaptor<T, U>::size() const noexcept
+    {
+        const double new_size = static_cast<double>(_buffer.size()) * _sizeof_ratio;
+        SPARROW_ASSERT(
+            std::trunc(new_size) == new_size,
+            "The size of the buffer is not a multiple of the size of the new type"
+        );
```
JohanMabille, line 303: Should we express this as a requirement on the size of `T` and `U` instead? I don't see a case where we would convert a type into another with a non integer size of ratio.

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,499 @@
...
+    constexpr bool buffer_adaptor<T, U>::empty() const noexcept
+    {
+        return _buffer.empty();
+    }
+
+    template <typename T, typename U>
+        requires(sizeof(U) < sizeof(T))
+    constexpr void buffer_adaptor<T, U>::reserve(size_type new_cap)
+    {
+        _buffer.reserve(static_cast<size_type>(static_cast<double>(new_cap) / _sizeof_ratio));
```
JohanMabille, line 332: Same comment as for `index_for_buffer` implementation.

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: If sizeof(U) > sizeof(T), for example U is 8 bytes and T is 1 bytes.
If you want to insert/resize 2 T elements, you have to insert 1 U element.
So after inserting 2 T elements, the size of the buffer_adaptor is now 8. 
BTW, we should also have the requirement that sizeof(T) i a multiple of sizeof(U).

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: Ok

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: Good idea !

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: 👍

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: Done

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: Good idea

include/sparrow/buffer.hpp
```diff
@@ -438,6 +443,15 @@ namespace sparrow
         get_data().p_end = copy_initialize(first, last, get_data().p_begin, get_allocator());
     }
 
+    template <class T>
+    template <class Range, allocator A>
```
Klaim, line 447: Consider short-cutting to `template<std::ranges::input_range Range, allocator A>`

include/sparrow/buffer.hpp
```diff
@@ -149,6 +150,10 @@ namespace sparrow
         template <class It, allocator A = allocator_type>
         constexpr buffer(It first, It last, const A& a = A());
 
+        template <class Range, allocator A = allocator_type>
```
Klaim, line 153: Consider short-cutting to `template<std::ranges::input_range Range, allocator A>`

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,503 @@
...
+#include <limits>
+#include <utility>
+
+#include "sparrow/buffer.hpp"
+#include "sparrow/contracts.hpp"
+
+namespace sparrow
+{
+    /**
+     * @brief Class which have internally a reference to a buffer<From> and provides the API of a buffer<To>
```
Klaim, line 27: Remember that we rely on autobrief

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,503 @@
...
+            "The size of the buffer is not a multiple of the size of the new type"
+        );
+        return static_cast<size_type>(new_size);
+    }
+
+    template <typename To, typename From>
+        requires(sizeof(From) < sizeof(To))
+    constexpr typename buffer_adaptor<To, From>::size_type buffer_adaptor<To, From>::max_size() const noexcept
+    {
+        return static_cast<size_type>(static_cast<double>(m_buffer.max_size()) * m_from_to_size_ratio);
```
Klaim, line 311: Could you isolate/factorize that calculation so that it's not repeated in `capacity()` etc.? From what I understand only the initial value is a paremeter.

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    template <typename To, typename From>
+        requires(sizeof(From) < sizeof(To))
+    constexpr typename buffer_adaptor<To, From>::iterator buffer_adaptor<To, From>::erase(const_iterator pos)
+    {
+        SPARROW_ASSERT_TRUE(cbegin() <= pos);
+        SPARROW_ASSERT_TRUE(pos <= cend());
+        const difference_type index = std::distance(cbegin(), pos);
+        const auto idx_for_buffer = index_for_buffer(static_cast<size_type>(index));
```
Klaim, line 429: When calling this function for immediate use in a buffer, consider adding assertions to check that index's bounds. (same in other functions)

test/test_buffer_adaptor.cpp
```diff
@@ -0,0 +1,610 @@
...
+#include "sparrow/buffer_adaptor.hpp"
+
+#include "doctest/doctest.h"
+
+namespace sparrow
+{
+
+    TEST_SUITE("buffer_adaptor")
+    {
+        const std::array<uint8_t, 8> input{1u, 2u, 3u, 4u, 5u, 6u, 7u, 8u};
```
Klaim, line 27: Also test with zero-sized array?

include/sparrow/buffer_adaptor.hpp
JohanMabille: Indeed, my assumption was that something stored as uint8_t and that we cast in another type would actually be stored as uint8_t for technical reasons (dynamic type, deserialization), and that such a case would never happen. But that mmight not be the case, and anyway, we want to catch this kind of bugs.

include/sparrow/buffer_adaptor.hpp
JohanMabille: Indeed, but I wonder if we should not add some constraints on the number of new elements we can insert / push_back etc in that case. That goes beyond our use case though (we will mostly convert from uint8_t to anything else), so let's keep it as is for now.

JohanMabille (approved):
LGTM once @Klaim  's comments have been addressed.

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: Done

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: I modified the code to avoid this long line of code

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: Done

include/sparrow/buffer.hpp
Alex-PLACET: Done

include/sparrow/buffer.hpp
Alex-PLACET: Done

include/sparrow/buffer_adaptor.hpp
```diff
@@ -0,0 +1,526 @@
...
+                                                       / static_cast<double>(sizeof(value_type));
+
+        static constexpr size_type m_to_from_size_ratio = static_cast<size_type>(1. / m_from_to_size_ratio);
+
+        [[nodiscard]] constexpr buffer_reference::size_type index_for_buffer(size_type idx) const noexcept;
+
+        [[nodiscard]] buffer_reference::const_iterator get_buffer_reference_iterator(const_iterator pos);
+
+        buffer_reference& m_buffer;
+        const size_type m_max_size;
```
Klaim, line 133: The `const` makes this type non-copyable/movable. Do we really want that?
If yes, then add copy/move operations explicitly deleted so that it's clear to the user.

Klaim: Note also the tradefoff here of using memory instead of quick calculation (from a function), I'm not sure what's the best for our users. But it's easy to change later if needed (we are not ABI-sensible).

Klaim (approved):
LGTM once the `const` member thing is clarified 👍🏽

include/sparrow/buffer_adaptor.hpp
Alex-PLACET: I made the member mutable 👍

PR #102, opened 2024-06-12
Sanitizers

Alex-PLACET:


.github/workflows/linux.yml
```diff
@@ -63,7 +64,7 @@ jobs:
         cache-downloads: true
 
     - name: Configure using CMake
-      run: cmake -G Ninja -Bbuild ${{matrix.sys.config-flags}} -DCMAKE_BUILD_TYPE:STRING=${{matrix.config.name}} -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DUSE_DATE_POLYFILL=${{matrix.sys.date-polyfill}} -DBUILD_TESTS=ON -DBUILD_EXAMPLES=ON
+      run: cmake -G Ninja -Bbuild ${{matrix.sys.config-flags}} -DCMAKE_BUILD_TYPE:STRING=${{matrix.config.name}} -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DUSE_DATE_POLYFILL=${{matrix.sys.date-polyfill}} -DBUILD_TESTS=ON -DBUILD_EXAMPLES=ON -DUSE_SANITIZER="address;undefined;leak"
```
Klaim, line 67: Note that sanitizers will change the runtime behavior of the program (not it's result) so it's best if it's run as a separate configuration. Maybe add a matrix variables to enable them?

Or do you think it's not worth separating?

.github/workflows/linux.yml
```diff
@@ -76,14 +77,20 @@ jobs:
     - name: Run tests
       working-directory: build
       run: cmake --build . --config ${{matrix.config.name}} --target run_tests_with_junit_report
+      env:
+        ASAN_OPTIONS: log_path=${{github.workspace}}/asan.log:alloc_dealloc_mismatch=0:halt_on_error=0:handle_abort=0:exitcode=0
```
Klaim, line 81: Could you add a comment clarifying why these  flags are necessary?

include/sparrow/allocator.hpp
```diff
@@ -53,6 +53,15 @@ namespace sparrow
                                     && (std::same_as<A, std::allocator<T>>
                                         || std::same_as<A, std::pmr::polymorphic_allocator<T>>);
 
+    template <class... Ts>
+    struct overloaded : Ts...
+    {
+        using Ts::operator()...;
+    };
+    // Although not required in C++20, clang needs it to build the code below
```
Klaim, line 61: Note that this code has been moved but the comemnts refers to "the code below".
1. Why was the move necessary?
2. Should the comment be updated?

.github/workflows/linux.yml
Alex-PLACET: Idk, if we run it in a separate configuration, we will double all jobs and reach 20 linux jobs.
As you said, the result should not change, so it's not really an issue for me.

include/sparrow/allocator.hpp
Alex-PLACET: The issue was that the `overloaded` definition must not be in a class definition (`any_allocator`).

.github/workflows/linux.yml
Klaim: Ok then lets go with this for now and change later. I suspect it might be problematic to not test exactly the same thing we deliver.

Klaim:
Could you add a commit adding a memory issue to check that the report works as intended? We'll drop the commit after we see the report.

JohanMabille (approved):

JohanMabille:
> Could you add a commit adding a memory issue to check that the report works as intended? We'll drop the commit after we see the report.

Was it done (and the commit has already been removed) ? If so, I think we can merge this PR.

Alex-PLACET:
It seems that we have an issue with the logs file. If you check the tests execution, you will see several leaks. But the log file is not create. Let me check that first.

PR #109, opened 2024-06-21
Add publish test report github action

Alex-PLACET:
This PR is to unlock the branch the improve the reporting of the UT reports as the workflow which depends on another workflow must be on the default branch.

### DIFF
--- a/.github/workflows/publish_test_report.yml
+++ b/.github/workflows/publish_test_report.yml
@@ -1,0 +1,20 @@
+name: 'Test Report'
+on:
+  workflow_run:
+    workflows: []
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_test_report.yml
```diff
@@ -0,0 +1,20 @@
+name: 'Test Report'
+on:
+  workflow_run:
+    workflows: []
```
Alex-PLACET, line 4: I will update the list of workflow in my next PR, currently it's ineffective

JohanMabille (approved):

PR #122, opened 2024-06-28
added docs via doxygen

DerThorsten:
* added Doxygen documentation
* using css from https://jothepro.github.io/doxygen-awesome-css/ (directly copied to the docs src)
* the main page and any other custom / hand written pages are generated from markdown (see docs/src/main_page.md)
* Added "examples" logic st. an example section can be added
* added typed_array example (@JohanMabille not sure if this is the intended usage, feel free to change the example code I added)
* run all examples in CI to ensure they are not broken 
* added ci workflow to: 
   * build the documentation as an artifact
   * upload the documentation to GitHub pages when changes to the main are made


DerThorsten:
The docs can be downloaded from the CI summary https://github.com/man-group/sparrow/actions/runs/9710061973?pr=122

JohanMabille (approved):
Creating a typed array will be way easier when we have finished the top level APIs, and we can eventually rewrite the exmaples. But let's keep it as is for now.

Issue #108, opened 2024-06-21
Clang OSX compilation issue


SylvainCorlay:
The compilation issue observed with standard clang (not Apple clang) on OSX actually applies to any build with clang with `libc++`.

It is due to this issue upstream: https://github.com/llvm/llvm-project/issues/67449, which has been resolved here: https://github.com/llvm/llvm-project/commit/078651b6de4b767b91e3e6a51e5df11a06d7bc4f#diff-cdad0245814da2f53716b70103381880154ed6ea30dcc34cfcc6c51a164c4b0d.

-- I am refering to the error message:

```
pointer_traits.h:118:22: error: implicit instantiation of undefined template 'std::__pointer_traits_element_type<S>'
  118 |     typedef typename __pointer_traits_element_type
  ```

PR #114, opened 2024-06-24
Add clangxx to OSX tests

SylvainCorlay:

PR #126, opened 2024-07-01
workaround for clang/apple/libc++

DerThorsten:
I see the same issue as mentioned in #108 

This crude workaround/hack allows me to compile on mac with clang

### DIFF
--- a/include/sparrow/typed_array.hpp
+++ b/include/sparrow/typed_array.hpp
@@ -467,7 +467,7 @@ namespace sparrow
     bool operator==(const typed_array<T, Layout>& ta1, const typed_array<T, Layout>& ta2)
     {   
         // see https://github.com/man-group/sparrow/issues/108
-#if defined(__APPLE__) && defined(__clang__) && defined(_LIBCPP_VERSION) && (_LIBCPP_VERSION < 180000)
+#if defined(_LIBCPP_VERSION) && (_LIBCPP_VERSION < 180000)
         if(ta1.size() != ta2.size())
         { 
             return false;
--- a/include/sparrow/typed_array.hpp
+++ b/include/sparrow/typed_array.hpp
@@ -467,8 +467,7 @@ namespace sparrow
     bool operator==(const typed_array<T, Layout>& ta1, const typed_array<T, Layout>& ta2)
     {   
         // see https://github.com/man-group/sparrow/issues/108
-#if defined(__APPLE__) && defined(__clang__) && defined(_LIBCPP_VERSION) && (__clang_major__ < 18)
-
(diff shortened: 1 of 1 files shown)


include/sparrow/typed_array.hpp
```diff
@@ -465,8 +465,25 @@ namespace sparrow
...
+        auto first2 = ta2.begin();
+        for(; first1 != last1; ++first1, ++first2){
+            if (!(*first1 == *first2)){
+                return false;
+            }
+        }
+        return true;
+#else
+        return std::equal(ta1.begin(), ta1.end(), ta2.begin(), ta2.end());
+#endif
```
SylvainCorlay, line 486: Thanks! I think that the std::equal should work with clang 18. Could we only do the workaround for clang < 18 ?

include/sparrow/typed_array.hpp
```diff
@@ -465,8 +465,25 @@ namespace sparrow
...
-    {
-        return std::equal(ta1.cbegin(), ta1.cend(), ta2.cbegin(), ta2.cend());
+    {   
+        // see https://github.com/man-group/sparrow/issues/108
+#if defined(__APPLE__) && defined(__clang__) && defined(_LIBCPP_VERSION)
+        if(ta1.size() != ta2.size())
+        { 
+            return false;
+        }
+        auto first1 = ta1.begin();
```
Klaim, line 475: Using `cbegin/cend` would be preferable here.

include/sparrow/typed_array.hpp
SylvainCorlay: Correct

DerThorsten: Any suggestions how to detect libc++<18?

chatgtp suggests some date based versioning shenanigans for LIBCPP
`#if defined(__APPLE__) && defined(__clang__) && defined(_LIBCPP_VERSION) && (_LIBCPP_VERSION < 20220701) `

jjerphan: Based on [the definition of `_LIBCPP_VERSION`](https://github.com/llvm/llvm-project/blob/main/libcxx/include/__config#L27-L30), this should work:

```c++
#if defined(__APPLE__) && defined(__clang__) && defined(_LIBCPP_VERSION) && (_LIBCPP_VERSION < 190000)
```

Klaim: My understanding is that `defined(__APPLE__) && defined(__clang__) && ` can then be removed. `libc++` can be used by other compilers than clang (like GCC and mingw, not sure about msvc) and (IMUIC) this is not specific to Apple platforms.

JohanMabille (approved):

Klaim (approved):
LGTM 👍🏽 in particular if it works with https://github.com/man-group/sparrow/pull/114

SylvainCorlay (approved):

PR #132, opened 2024-07-03
make  const ref proxy assignable to  ref proxy

DerThorsten:


JohanMabille:
I am currently rewriting these classes, we can close this kne.

PR #135, opened 2024-07-03
sparrow::nullable implementation

JohanMabille:
This class is meant to replace the `reference_proxy` and `const_reference_proxy` classes defined in array_data.hpp. It's still missing:
- ~~comparison operators~~
- ~~tests~~
- ~~documentation~~

### DIFF
--- a/include/sparrow/mp_utils.hpp
+++ b/include/sparrow/mp_utils.hpp
@@ -419,4 +419,8 @@ namespace sparrow::mpl
 #endif
     }
 
+    /// The boolean_like concept specifies that a type can be convertible to and assignable from
+    /// bool.
+    template <class T>
+    concept boolean_like = std::constructible_from<T, bool> and std::convertible_to<T, bool>;
 }
--- a/include/sparrow/mp_utils.hpp
+++ b/include/sparrow/mp_utils.hpp
@@ -419,8 +419,9 @@ namespace sparrow::mpl
 #endif
     }
 
-    /// The boolean_like concept specifies that a type can be convertible to and assignable from
-    /// bool.
+    /// Matches types that can be convertible to and assignable from bool. We do not use
+    /// `std::convertible_to` because we don't want to impose an implicit conversion.
     template <class T>
-    concept boolean_like = std::is_assignable_v<std::add_lvalue_reference_t<T>, bool> and std::convertible_to<T, bool>;
+    concept boolean_like = std::is_assignable_v<std::add_lvalue_reference_t<T>, bool> and 
+                           requires { static_cast<bool>(std::declval<T>()); };
 }
--- a/include/sparrow/mp_utils.hpp
+++ b/include/sparrow/mp_utils.hpp
@@ -422,5 +422,5 @@ namespace sparrow::mpl
     /// The boolean_like concept specifies that a type can be convertible to and assignable from
     /// bool.
     template <class T>
-    concept boolean_like = std::constructible_from<T, bool> and std::convertible_to<T, bool>;
+    concept boolean_like = std::is_assignable_v<std::add_lvalue_reference_t<T>, bool> and std::convertible_to<T, bool>;
 }
--- a/include/sparrow/mp_utils.hpp
+++ b/include/sparrow/mp_utils.hpp
@@ -355,6 +355,15 @@ namespace sparrow::mpl
     //////////////////////////////////////////////////
     //// Miscellaneous ///////////////////////////////
 
+    template <class T>
+    struct add_const_lvalue_reference
+        : std::add_lvalue_reference<std::add_const_t<T>>
+    {
+    };
+
+    template <class T>
+    using add_const_lvalue_reference_t = typename add_const_lvalue_reference<T>::type;
+
     template <class T, bool is_const>
     struct constify : std::conditional<is_const, const T, T>
     {
--- a/include/sparrow/nullable.hpp
+++ b/include/sparrow/nullable.hpp
@@ -83,7 +83,7 @@ namespace sparrow
 
     private:
 
-        static constexpr const char* message = "bad nullable access";
+        static constexpr const char* message = "Invalid access to null value";
     };
 
     /**
@@ -91,6 +91,11 @@ namespace sparrow
      */
     struct nullval_t
     {
+        // This is required to disable the generation of
+        // the default constructor. Otherwise, a = {} where
+        // a is a nullable would lead to an ambiguous call
+        // where both operator=(nullable&&) and operator=(nullval_t)
+        // are valid.
         constexpr explicit nullval_t(int) {}
     };
 
@@ -128,9 +133,12 @@ namespace sparrow
         concept initializable_from_refs =
             initializable_from_one<T, Arg&, const Arg&, Arg&&, const Arg&&>;
 
+        // We prefer std::is_assignable_v to std::assignable_from<To, From> because
+        // std::assignable_from requires the existence of an implicit conversion
+        // from From to To
         template <class To, class... Args>
         concept assignable_from_one =
-            (std::assignable_from<std::add_lvalue_reference_t<To>, Args> && ...);
+            (std::is_assignable_v<std::add_lvalue_reference<To>, Args> && ...);
 
         template <class T, class Arg>
         concept assignable_from_refs =
@@ -151,20 +159,25 @@ namespace sparrow
 
         template <class To1, class From1, class To2, class From2>
         concept both_assignable_from_cref =
-            std::assignable_from<std::add_lvalue_reference_t<To1>, mpl::add_const_lvalue_reference_t<From1>> and
-            std::assignable_from<std::add_lvalue_reference_t<To2>, mpl::add_const_lvalue_reference_t<From2>&>;
+            std::is_assignable_v<std::add_lvalue_reference_t<To1>, mpl::add_const_lvalue_reference_t<From1>> and
+            std::is_assignable_v<std::add_lvalue_reference_t<To2>, mpl::add_const_lvalue_reference_t<From2>>;
 
         template <class To1, class From1, class To2, class From2>
         concept both_assignable_from_cond_ref =
-            std::assignable_from<std::add_lvalue_reference_t<To1>, conditional_ref_t<To1, From1>> and
-            std::assignable_from<std::add_lvalue_reference_t<To2>, conditional_ref_t<To2, From2>>;
+            std::is_assignable_v<std::add_lvalue_reference_t<To1>, conditional_ref_t<To1, From1>> and
+            std::is_assignable_v<std::add_lvalue_reference_t<To2>, conditional_ref_t<To2, From2>>;
 
         template <class T>
         static constexpr bool is_nullable_v = mpl::is_type_instance_of_v<T, nullable>;
     }
 
     /**
-     * The nullable class.
+     * The nullable class models a value or a reference that can be "null", or missing.
+     * The flag indicating whether the element should be considered missing can be a
+     * boolean-like value or reference.
+     *
+     * The value is always available, independently from the value of the flag. The flag
+     * only indicates whether the value should be considered for computation.
      */
     template <class T, boolean_like B = bool>
     class nullable
@@ -185,15 +198,17 @@ namespace sparrow
         using flag_rvalue_reference = typename flag_traits::rvalue_reference;
         using flag_const_rvalue_reference = typename flag_traits::const_rvalue_reference;
         
+        template <std::default_initializable U = T, std::default_initializable BB = B>
         constexpr nullable() noexcept
             : m_value()
-            , m_flag(false)
+            , m_null_flag(false)
         {
         }
         
+        template <std::default_initializable U = T, std::default_initializable BB = B>
         constexpr nullable(nullval_t) noexcept
             : m_value()
-            , m_flag(false)
+            , m_null_flag(false)
         {
         }
         
@@ -205,7 +220,7 @@ namespace sparrow
         explicit (not std::convertible_to<U&&, T>)
         constexpr nullable(U&& value)
             : m_value(std::forward<U>(value))
-            , m_flag(true)
+            , m_null_flag(true)
         {
         }
 
@@ -219,7 +234,7 @@ namespace sparrow
         explicit(not impl::both_convertible_from_cref<T, TO, B, BO>)
         constexpr nullable(const nullable<TO, BO>& rhs)
             : m_value(rhs.get())
-            , m_flag(rhs.has_value())
+            , m_null_flag(rhs.null_flag())
         {
         }
 
@@ -233,37 +248,37 @@ namespace sparrow
         explicit(not impl::both_convertible_from_cond_ref<T, TO, B, BO>)
         constexpr nullable(nullable<TO, BO>&& rhs)
             : m_value(std::move(rhs).get())
-            , m_flag(std::move(rhs).has_value())
+            , m_null_flag(std::move(rhs).null_flag())
         {
         }
 
-        constexpr nullable(value_type&& value, flag_type&& flag)
+        constexpr nullable(value_type&& value, flag_type&& null_flag)
             : m_value(std::move(value))
-            , m_flag(std::move(flag))
+            , m_null_flag(std::move(null_flag))
         {
         }
 
-        constexpr nullable(std::add_lvalue_reference_t<T> value, std::add_lvalue_reference_t<B> flag)
+        constexpr nullable(std::add_lvalue_reference_t<T> value, std::add_lvalue_reference_t<B> null_flag)
             : m_value(value)
-            , m_flag(flag)
+            , m_null_flag(null_flag)
         {
         }
 
-        constexpr nullable(value_type&& value, std::add_lvalue_reference_t<B> flag)
+        constexpr nullable(value_type&& value, std::add_lvalue_reference_t<B> null_flag)
             : m_value(std::move(value))
-            , m_flag(flag)
+            , m_null_flag(null_flag)
         {
         }
 
-        constexpr nullable(std::add_lvalue_reference_t<T> value, flag_type&& flag)
+        constexpr nullable(std::add_lvalue_reference_t<T> value, flag_type&& null_flag)
             : m_value(value)
-            , m_flag(std::move(flag))
+            , m_null_flag(std::move(null_flag))
         {
         }
 
         constexpr self_type& operator=(nullval_t)
         {
-            m_flag = false;
+            m_null_flag = false;
             return *this;
         }
 
@@ -275,14 +290,14 @@ namespace sparrow
         constexpr self_type& operator=(TO&& rhs)
         {
             m_value = std::forward<TO>(rhs);
-            m_flag = true;
+            m_null_flag = true;
             return *this;
         }
 
         constexpr self_type& operator=(const self_type& rhs)
         {
             m_value = rhs.get();
-            m_flag = rhs.has_value();
+            m_null_flag = rhs.null_flag();
             return *this;
         }
 
@@ -295,14 +310,14 @@ namespace sparrow
         constexpr self_type& operator=(const nullable<TO, BO>& rhs)
         {
             m_value = rhs.get();
-            m_flag = rhs.has_value();
+            m_null_flag = rhs.null_flag();
             return *this;
         }
 
         constexpr self_type& operator=(self_type&& rhs)
         {
             m_value = std::move(rhs).get();
-            m_flag = std::move(rhs).has_value();
+            m_null_flag = std::move(rhs).null_flag();
             return *this;
         }
 
@@ -315,16 +330,17 @@ namespace sparrow
         constexpr self_type& operator=(nullable<TO, BO>&& rhs)
         {
             m_value = std::move(rhs).get();
-            m_flag = std::move(rhs).has_value();
+            m_null_flag = std::move(rhs).null_flag();
             return *this;
         }
 
         constexpr explicit operator bool() const noexcept;
+        constexpr bool has_value() const noexcept;
 
-        constexpr flag_reference has_value() & noexcept;
-        constexpr flag_const_reference has_value() const & noexcept;
-        constexpr flag_rvalue_reference has_value() && noexcept;
-        constexpr flag_const_rvalue_reference has_value() const && noexcept;
+        constexpr flag_reference null_flag() & noexcept;
+        constexpr flag_const_reference null_flag() const & noexcept;
+        constexpr flag_rvalue_reference null_flag() && noexcept;
+        constexpr flag_const_rvalue_reference null_flag() const && noexcept;
 
         constexpr reference get() & noexcept;
         constexpr const_reference get() const & noexcept;
@@ -350,7 +366,7 @@ namespace sparrow
         void throw_if_null() const;
 
         T m_value;
-        B m_flag;
+        B m_null_flag;
 
         template <class TO, boolean_like BO>
         friend class nullable;
@@ -380,9 +396,6 @@ namespace sparrow
     constexpr std::compare_three_way_result_t<T, U>
     operator<=>(const nullable<T, B>& lhs, const nullable<U, UB>& rhs) noexcept;
 
-    template <class T, boolean_like B = bool>
-    constexpr nullable<T, B> make_nullable(T&& value, B&& flag = true);
-
     /***************************
      * nullable implementation *
      ***************************/
@@ -390,44 +403,50 @@ namespace sparrow
     template <class T, boolean_like B>
     constexpr nullable<T, B>::operator bool() const noexcept
     {
-        return m_flag;
+        return m_null_flag;
     }
 
     template <class T, boolean_like B>
-    constexpr auto nullable<T, B>::has_value() & noexcept -> flag_reference
+    constexpr bool nullable<T, B>::has_value() const noexcept
     {
-        return m_flag;
+        return m_null_flag;
+    }
+
+    template <class T, boolean_like B>
+    constexpr auto nullable<T, B>::null_flag() & noexcept -> flag_reference
+    {
+        return m_null_flag;
     }
     
     template <class T, boolean_like B>
-    constexpr auto nullable<T, B>::has_value() const & noexcept -> flag_const_reference
+    constexpr auto nullable<T, B>::null_flag() const & noexcept -> flag_const_reference
     {
-        return m_flag;
+        return m_null_flag;
     }
     
     template <class T, boolean_like B>
-    constexpr auto nullable<T, B>::has_value() && noexcept -> flag_rvalue_reference
+    constexpr auto nullable<T, B>::null_flag() && noexcept -> flag_rvalue_reference
     {
         if constexpr (std::is_reference_v<B>)
         {
-            return m_flag;
+            return m_null_flag;
         }
         else
         {
-            return flag_rvalue_reference(m_flag);
+            return flag_rvalue_reference(m_null_flag);
         }
     }
     
     template <class T, boolean_like B>
-    constexpr auto nullable<T, B>::has_value() const && noexcept -> flag_const_rvalue_reference
+    constexpr auto nullable<T, B>::null_flag() const && noexcept -> flag_const_rvalue_reference
     {
         if constexpr (std::is_reference_v<B>)
         {
-            return m_flag;
+            return m_null_flag;
         }
         else
         {
-            return flag_const_rvalue_reference(m_flag);
+            return flag_const_rvalue_reference(m_null_flag);
         }
     }
 
@@ -516,19 +535,19 @@ namespace sparrow
     {
         using std::swap;
         swap(m_value, other.m_value);
-        swap(m_flag, other.m_flag);
+        swap(m_null_flag, other.m_null_flag);
     }
 
     template <class T, boolean_like B>
     void nullable<T, B>::reset() noexcept
     {
-        m_flag = false;
+        m_null_flag = false;
     }
     
     template <class T, boolean_like B>
     void nullable<T, B>::throw_if_null() const
     {
-        if (!m_flag)
+        if (!m_null_flag)
         {
             throw bad_nullable_access{};
         }
@@ -577,12 +596,6 @@ namespace sparrow
     operator<=>(const nullable<T, B>& lhs, const nullable<U, UB>& rhs) noexcept
     {
         return (lhs && rhs) ? lhs.get() <=> rhs.get() : bool(lhs) <=> bool(rhs);
-    }
-
-    template <class T, boolean_like B>
-    constexpr nullable<T, B> make_nullable(T&& value, B&& flag)
-    {
-        return nullable<T, B>(std::forward<T>(value), std::forward<B>(flag));
     }
 
 }
--- a/include/sparrow/nullable.hpp
+++ b/include/sparrow/nullable.hpp
@@ -443,6 +443,11 @@ namespace sparrow
     constexpr std::compare_three_way_result_t<T, U>
     operator<=>(const nullable<T, B>& lhs, const nullable<U, UB>& rhs) noexcept;
 
+    // Even if we have CTAD in C++20, some constructors add lvalue reference
+    // to their argument, making the deduction impossible.
+    template <class T, boolean_like B = bool>
+    constexpr nullable<T, B> make_nullable(T&& value, B&& flag = true);
+    
     /***************************
      * nullable implementation *
      ***************************/
@@ -645,5 +650,10 @@ namespace sparrow
         return (lhs && rhs) ? lhs.get() <=> rhs.get() : bool(lhs) <=> bool(rhs);
     }
 
+    template <class T, boolean_like B>
+    constexpr nullable<T, B> make_nullable(T&& value, B&& flag)
+    {
+        return nullable<T, B>(std::forward<T>(value), std::forward<B>(flag));
+    }
 }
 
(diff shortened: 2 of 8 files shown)


include/sparrow/optional.hpp
Alex-PLACET, line 1: formatting :)

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+namespace sparrow
+{
+    using optional_double = optional<double, bool>;
+    using optional_int = optional<int, bool>;
+
+    TEST_SUITE("optional value")
+    {
+        TEST_CASE("constructors")
+        {
+            {
```
Alex-PLACET, line 28: NOt mandatory but you can use SUBCASE for each test where you use different constructors

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK(d.has_value());
+                CHECK_EQ(d.value(), 3.);
+            }
+
+            {
+                double val = 1.2;
+                bool b1 = true;
+
+                optional_double td1(val, b1);
+                optional_double td2(std::move(val), b1);
```
Alex-PLACET, line 56: The std::move is useless as the val is not movable. Not sure about what you want to test

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+            }
+
+            {
+                optional_double d(std::nullopt);
+                CHECK_FALSE(d.has_value());
+            }
+
+            {
+                optional_double d(1.2);
+                CHECK(d.has_value());
```
Alex-PLACET, line 40:
```suggestion
                REQUIRE(d.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK_EQ(d.value(), 1.2);
+            }
+
+            {
+                int i = 3;
+                optional_double d(i);
+                CHECK(d.has_value());
```
Alex-PLACET, line 47:
```suggestion
                REQUIRE(d.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                optional_double td4(std::move(val), std::move(b1));
+
+                CHECK(td1.has_value());
+                CHECK_EQ(td1.value(), val);
+                CHECK(td2.has_value());
+                CHECK_EQ(td2.value(), val);
+                CHECK(td3.has_value());
+                CHECK_EQ(td3.value(), val);
+                CHECK(td4.has_value());
+                CHECK_EQ(td4.value(), val);
```
Alex-PLACET, line 67:
```suggestion
                REQUIRE(td1.has_value());
                CHECK_EQ(td1.value(), val);
                REQUIRE(td2.has_value());
                CHECK_EQ(td2.value(), val);
                REQUIRE(td3.has_value());
                CHECK_EQ(td3.value(), val);
                REQUIRE(td4.has_value());
                CHECK_EQ(td4.value(), val);
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK_FALSE(fd4.has_value());
+            }
+        }
+
+        TEST_CASE("copy constructors")
+        {
+            {
+                optional_double d1(1.2);
+                optional_double d2(d1);
+                CHECK(d2.has_value());
```
Alex-PLACET, line 87:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK_EQ(d1.value(), d2.value());
+            }
+
+            {
+                optional_int i(2);
+                optional_double d(i);
+                CHECK(d.has_value());
```
Alex-PLACET, line 94:
```suggestion
                REQUIRE(d.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+            }
+        }
+
+        TEST_CASE("move constructors")
+        {
+            {
+                optional_double d0(1.2);
+                optional_double d1(d0);
+                optional_double d2(std::move(d0));
+                CHECK(d2.has_value());
```
Alex-PLACET, line 111:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK_EQ(d1.value(), d2.value());
+            }
+            {
+                optional_int i(2);
+                optional_int ci(i);
+                optional_double d(std::move(i));
+                CHECK(d.has_value());
```
Alex-PLACET, line 118:
```suggestion
                REQUIRE(d.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+            }
+        }
+
+        TEST_CASE("copy assign")
+        {
+            {
+                optional_double d1(1.2);
+                optional_double d2(2.5);
+                d2 = d1;
+                CHECK(d2.has_value());
```
Alex-PLACET, line 134:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK_EQ(d1.value(), d2.value());
+            }
+            {
+                optional_int d1(1);
+                optional_double d2(2.5);
+                d2 = d1;
+                CHECK(d2.has_value());
```
Alex-PLACET, line 141:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+        }
+
+        TEST_CASE("move assign")
+        {
+            {
+                optional_double d0(1.2);
+                optional_double d1(d0);
+                optional_double d2(2.5);
+                d2 = std::move(d0);
+                CHECK(d2.has_value());
```
Alex-PLACET, line 159:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK_EQ(d1.value(), d2.value());
+            }
+            {
+                optional_int d0(1);
+                optional_int d1(d0);
+                optional_double d2(2.5);
+                d2 = std::move(d0);
+                CHECK(d2.has_value());
```
Alex-PLACET, line 167:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+    TEST_SUITE("optional proxy")
+    {
+        TEST_CASE("constructors")
+        {
+            double val = 1.2;
+            bool b1 = true;
+
+            optional_proxy td(val);
+            CHECK(td.has_value());
```
Alex-PLACET, line 355:
```suggestion
            REQUIRE(td.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+            CHECK_EQ(td.value(), val);
+
+            optional_proxy td1(val, b1);
+            optional_proxy td2(val, std::move(b1));
+
+            CHECK(td1.has_value());
+            CHECK_EQ(td1.value(), val);
+            CHECK(td2.has_value());
+            CHECK_EQ(td2.value(), val);
```
Alex-PLACET, line 364:
```suggestion
            REQUIRE(td1.has_value());
            CHECK_EQ(td1.value(), val);
            REQUIRE(td2.has_value());
            CHECK_EQ(td2.value(), val);
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+            CHECK_FALSE(fd1.has_value());
+            CHECK_FALSE(fd2.has_value());
+        }
+
+        TEST_CASE("copy constructors")
+        {
+            double val = 1.2;
+            optional_proxy d1(val);
+            optional_proxy d2(d1);
+            CHECK(d2.has_value());
```
Alex-PLACET, line 379:
```suggestion
            REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+            CHECK_EQ(d1.value(), d2.value());
+        }
+
+        TEST_CASE("move constructor")
+        {
+            double val = 1.2;
+            optional_proxy d1(val);
+            optional_proxy d2(std::move(d1));
+            CHECK(d2.has_value());
```
Alex-PLACET, line 388:
```suggestion
            REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+        TEST_CASE("copy assign")
+        {
+            {
+                double initial = 1.2;
+                double expected = 2.5;
+                optional_proxy d1(initial);
+                optional_proxy d2(expected);
+                d2 = d1;
+                CHECK(d2.has_value());
```
Alex-PLACET, line 400:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+                CHECK_EQ(initial, expected);
+            }
+
+            {
+                double initial = 1.2;
+                double expected = 2.5;
+                optional_double d1(initial);
+                optional_proxy d2(expected);
+                d2 = d1;
+                CHECK(d2.has_value());
```
Alex-PLACET, line 411:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+        TEST_CASE("move assign")
+        {
+            {
+                double initial = 1.2;
+                double expected = 2.5;
+                optional_proxy d1(initial);
+                optional_proxy d2(expected);
+                d2 = std::move(d1);
+                CHECK(d2.has_value());
```
Alex-PLACET, line 432:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+            }
+
+            {
+                double initial = 1.2;
+                double expected = 2.5;
+                optional_double d1(initial);
+                optional_proxy d2(expected);
+                d2 = std::move(d1);
+                CHECK(d2.has_value());
```
Alex-PLACET, line 444:
```suggestion
                REQUIRE(d2.has_value());
```

test/test_optional.cpp
JohanMabille: So here we want to the the constructor overload that takes an rvalue reference. `std::move` does not actually "move" anything, it just casts its argument to an rvalue reference. Whether we have a moving constructor called in the end or not is not what we are intereested in here.

include/sparrow/optional.hpp
```diff
@@ -121,6 +121,14 @@ namespace sparrow
         static constexpr bool is_optional_v = is_optional<T>::value;
     }
 
+    /**
+     * The optional class is similar to std::optional, with two major differences:
+     * - it can act as a proxy, meaning its template parameter can be lvalue references
+     *   or lvalue const references
+     * - the semantic for empty optoinal: resetting an non empty optional does not destruct
```
Alex-PLACET, line 128:
```suggestion
     * - the semantic for empty optional: resetting an non empty optional does not destruct
```

include/sparrow/optional.hpp
```diff
@@ -121,6 +121,14 @@ namespace sparrow
...
+     *   the contained value. Allocating an empty optional construct the contained value. This
+     *   behavior is temporary and will be changed in the near future.
+     */
     template <class T, class B = bool>
     class optional
```
Alex-PLACET, line 133:
```suggestion
    template <class T, class B = bool>
        requires std::constructible_from<B, bool>
    class optional
```

JohanMabille: It also requires to be convertible to bool Let's create a concept and use it in front of the template parameter, otherwise we need to implement the class in its definition.

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+    {
+        throw_if_empty();
+        return std::move(m_value);
+    }
+
+    template <class T, boolean_like B>
+    template <class U>
+    constexpr auto optional<T, B>::value_or(U&& default_value) const & -> value_type
+    {
+        return bool(*this) ? **this : static_cast<value_type>(std::forward<U>(default_value));
```
Klaim, line 415: - `bool(*this)` the cast should not be necessary?
- `**this` I would prefer an explicit `value()` being called here.
- I would prefer the `static_cast` to be replaced by a construction (note that there is no concepts here like in the constructors, maybe there should be)

same issues below

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+    {
+        return bool(*this) ? std::move(**this) : static_cast<value_type>(std::forward<U>(default_value)); 
+    }
+
+    template <class T, boolean_like B>
+    void optional<T, B>::swap(self_type& other) noexcept
+    {
+        using std::swap;
+        swap(m_value, other.m_value);
+        swap(m_flag, other.m_flag);
```
Klaim, line 430: Notice the same `m_flag = false` issue mentionned before

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+    template <class T, class B>
+    constexpr bool operator==(const optional<T, B>& lhs, std::nullopt_t) noexcept
+    {
+        return !lhs;
+    }
+
+    template <class T, class B>
+    constexpr std::strong_ordering operator<=>(const optional<T, B>& lhs, std::nullopt_t) noexcept
+    {
+        return bool(lhs) <=> false;
```
Klaim, line 463: Why is the cast necessary?

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+    }
+    
+    template <class T, class B, class U>
+    constexpr bool operator==(const optional<T, B>& lhs, const U& rhs) noexcept
+    {
+        return bool(lhs) && *lhs == rhs;
```
Klaim, line 469: Add parenthesis for clarity of the order of evaluation

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+    }
+
+    template <class T, class B, class U>
+    requires (!util::is_optional_v<U> && std::three_way_comparable_with<U, T>)
+    constexpr std::compare_three_way_result_t<T, U>
+    operator<=>(const optional<T, B>& lhs, const U& rhs) noexcept
+    {
+        return bool(lhs) ? *lhs <=> rhs : std::strong_ordering::less;
```
Klaim, line 477: I dont think the cast is necessary either here

test/test_optional.cpp
```diff
@@ -0,0 +1,697 @@
...
+        {
+            SUBCASE("default")
+            {
+                optional_double d;
+                CHECK_FALSE(d.has_value());
+            }
+            
+            SUBCASE("from nullopt")
+            {
+                optional_double d(std::nullopt);
```
Klaim, line 36: Prefer `{std::nullopt}` instead (and in general, where there is no `initializer_list` involved), we want to avoid situations where an implicit cast happen and we didnt realize.

test/test_optional.cpp
```diff
@@ -0,0 +1,697 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "sparrow/optional.hpp"
+
+#include "doctest/doctest.h"
+
+namespace sparrow
+{
+    using optional_double = optional<double, bool>;
```
Klaim, line 21: Missing tests I would like to see:
- find a way to have all the tests in `consteval` context, to make sure the `constexpr` in the class definition all work compile-time (except maybe for the custom type below); Note that code that can be tested at compile-time is guaranteed to not compile if there is UB situations in it, so that's a big win.
- parameterize the tests so that we can check with other type than `double` and `int`
- set T as a custom type which set a global boolean on construction and destruction
  - it must not be set instantiated when default-construction the optional
  - it must be destroyed when the boolean is set to false

test/test_optional.cpp
```diff
@@ -0,0 +1,697 @@
...
+    using optional_proxy = optional<double&, bool>;
+
+    TEST_SUITE("optional proxy")
+    {
+        TEST_CASE("constructors")
+        {
+            double val = 1.2;
+            bool b1 = true;
+
+            optional_proxy td(val);
```
Klaim, line 398: Prefer `{`

test/test_optional.cpp
```diff
@@ -0,0 +1,697 @@
...
+        }
+    }
+
+    using optional_proxy = optional<double&, bool>;
+
+    TEST_SUITE("optional proxy")
+    {
+        TEST_CASE("constructors")
+        {
+            double val = 1.2;
```
Klaim, line 395: Set all the constants evaluated against as `const` (in all tests that have them)

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+            std::constructible_from<T, mpl::add_const_lvalue_reference_t<TArgs>> and
+            std::constructible_from<U, mpl::add_const_lvalue_reference_t<UArgs>>;
+
+        template <class To1, class From1, class To2, class From2>
+        concept both_convertible_from_cref =
+            std::convertible_to<mpl::add_const_lvalue_reference_t<From1>, To1> and
+            std::convertible_to<mpl::add_const_lvalue_reference_t<From2>, To2>;
+
+        template <class T, class... Args>
+        concept constructible_from_one =
```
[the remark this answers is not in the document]
JohanMabille: `constructible_from_each` would mean that T can be constructed form every Args; while `constructible_from_one` means it has to be constructible from at least one argument in `Args`.

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+        {
+        }
+        
+        template <class U>
+        requires (
+            not std::same_as<self_type, std::decay_t<U>> and
+            std::constructible_from<T, U&&>
+        )
+        explicit (not std::convertible_to<U&&, T>)
+        constexpr optional(U&& value)
```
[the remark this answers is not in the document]
JohanMabille: Indeed, I'll add it.

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+
+        constexpr optional(std::add_lvalue_reference_t<T> value, flag_type&& flag)
+            : m_value(value)
+            , m_flag(std::move(flag))
+        {
+        }
+
+        constexpr self_type& operator=(std::nullopt_t)
+        {
+            m_flag = false;
```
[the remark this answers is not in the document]
JohanMabille: Following in person discussion:
- this class is actualy badly named as it is a very different concept from optional. Nullable would be a better name.
- The value should always exist, even when the object is "nullified"

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+            not util::assignable_from_refs<T, optional<TO, BO>>
+        )
+        constexpr self_type& operator=(optional<TO, BO>&& rhs)
+        {
+            m_value = std::move(rhs.m_value);
+            m_flag = std::move(rhs.m_flag);
+            return *this;
+        }
+
+        constexpr bool has_value() const noexcept;
```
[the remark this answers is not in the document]
JohanMabille: I prefer to keep the declaration and implementation separated, so that it's easier to read the whole interface of a class. However, I wanted to avoid repeating the constraints of the different constructors, it would have been cumbersome.

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+        constexpr value_type value_or(U&& default_value) &&;
+
+        void swap(self_type& other) noexcept;
+        void reset() noexcept;
+
+    private:
+
+        void throw_if_empty() const;
+
+        T m_value;
```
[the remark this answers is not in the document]
JohanMabille: Moving to a nullable class where the value is always available solves this issue.

include/sparrow/optional.hpp
```diff
@@ -0,0 +1,500 @@
...
+    constexpr optional<T, B> make_optional(T&& value, B&& flag = true);
+
+    /***************************
+     * optional implementation *
+     ***************************/
+
+    template <class T, boolean_like B>
+    constexpr bool optional<T, B>::has_value() const noexcept
+    {
+        return m_flag;
```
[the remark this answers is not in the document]
JohanMabille: `boolea_like` forces the type to expose this conversion operator.

SylvainCorlay:
> the "empty semantic", i.e. do not call value type constructor when initializing to nullopt, and call value type destructor when resetting. This is not trivial to do since it does not totally apply to optional on references.

Commenting on this. Since we don't adhere to the std::optional semantics I don't think that we should do this. Basically, assigning a `sparrow::null` value to a `sparrow::nullable` should not call the destructor on the underlying value when it is owned by the nullable. I think it should only be deleted upon deletion of the nullable.

SylvainCorlay (approved):

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,589 @@
...
+        using flag_reference = typename flag_traits::reference;
+        using flag_const_reference = typename flag_traits::const_reference;
+        using flag_rvalue_reference = typename flag_traits::rvalue_reference;
+        using flag_const_rvalue_reference = typename flag_traits::const_rvalue_reference;
+        
+        constexpr nullable() noexcept
+            : m_value()
+            , m_flag(false)
+        {
+        }
```
SylvainCorlay, line 192: We should only create the default constructor -  and make nullable default-constructible if `T` and `B` are default constructible.

If they are not, the std::variant will also not be. This may solve the issue with the follow up PR.

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,589 @@
...
+
+        constexpr explicit operator bool() const noexcept;
+
+        constexpr flag_reference has_value() & noexcept;
+        constexpr flag_const_reference has_value() const & noexcept;
+        constexpr flag_rvalue_reference has_value() && noexcept;
+        constexpr flag_const_rvalue_reference has_value() const && noexcept;
+
+        constexpr reference get() & noexcept;
```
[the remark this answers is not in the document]
JohanMabille: There is always a value, whatever the flag is. The fls just means we should not consider it.

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,589 @@
...
+        constexpr self_type& operator=(nullable<TO, BO>&& rhs)
+        {
+            m_value = std::move(rhs).get();
+            m_flag = std::move(rhs).has_value();
+            return *this;
+        }
+
+        constexpr explicit operator bool() const noexcept;
+
+        constexpr flag_reference has_value() & noexcept;
```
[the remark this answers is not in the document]
JohanMabille: Having a way to change the flag only is harmless since there is no associated logic for destructing the value. Besides, these methods are required to correclty implement the different constructors.

However, I agree that the name is misleading. We could have a const has_value returning a bool, and rename these functions into "null_flag" for instance;

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,589 @@
...
+
+    template <class T, class B, class U, class UB>
+    constexpr bool operator==(const nullable<T, B>& lhs, const nullable<U, UB>& rhs) noexcept;
+
+    template <class T, class B, std::three_way_comparable_with<T> U, class UB>
+    constexpr std::compare_three_way_result_t<T, U>
+    operator<=>(const nullable<T, B>& lhs, const nullable<U, UB>& rhs) noexcept;
+
+    template <class T, boolean_like B = bool>
+    constexpr nullable<T, B> make_nullable(T&& value, B&& flag = true);
```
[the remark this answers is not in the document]
JohanMabille: I always forget that we have CTAD now ....

test/test_nullable.cpp
```diff
@@ -0,0 +1,697 @@
...
+            }
+
+            SUBCASE("from value")
+            {
+                nullable_double d{1.2};
+                REQUIRE(d.has_value());
+                CHECK_EQ(d.value(), 1.2);
+            }
+
+            SUBCASE("from value with conversion")
```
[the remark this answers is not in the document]
JohanMabille: That is what you have with `std::optional` for instance, or when you do arithmetics. You want to be able to mix int and double, and initialize an `std::optional<double>` from an int. The same applies to nullable, and we limit the conversions to those of the underlying types (just like std::optional does). Notice that we have the same logic in xtensor with the [xoptional](https://github.com/xtensor-stack/xtl/blob/master/include/xtl/xoptional.hpp) type (really similar to nullable, and badly named as the first implementation here ^^), and it has worked like a charm for many years.

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,589 @@
...
+
+        static constexpr const char* message = "bad nullable access";
+    };
+
+    /**
+     * nullval_t is an empty class used to indicate that a nullable is null.
+     */
+    struct nullval_t
+    {
+        constexpr explicit nullval_t(int) {}
```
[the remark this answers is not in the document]
JohanMabille: Otherwise `opt = {}` would lead to an ambiguity between `operator=(nullable&&)` and `operator=(nullval_t)`.

include/sparrow/nullable.hpp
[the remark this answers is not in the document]
Klaim: That would make more sense indeed 👍🏽 

> Having a way to change the flag only is harmless since there is no associated logic for destructing the value. 

That's assuming we will not need any other kind of code in there in the future, I wouldn't try to guess myself haha

> Besides, these methods are required to correclty implement the different constructors.

Could you clarify this point? Arent the constructors already able to access the private members anyway?

include/sparrow/nullable.hpp
JohanMabille: Notice this does not fix the followup PR, but we still need to do it. Otherwise, `std::default_constructible<nullable<double&, bool>>` is `true`, even if trying to default construct it leads to a compilation error.

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,589 @@
...
+
+    private:
+
+        static constexpr const char* message = "bad nullable access";
+    };
+
+    /**
+     * nullval_t is an empty class used to indicate that a nullable is null.
+     */
+    struct nullval_t
```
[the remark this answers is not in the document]
Klaim: I was thinking more about [how `std::byte` was defined](https://en.cppreference.com/w/cpp/types/byte) (no need to access a value, there is none), but either is fine to me, I'm not sure it would help here.

include/sparrow/nullable.hpp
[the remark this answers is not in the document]
Klaim: Could you add this explaination as a comment next to that line?

include/sparrow/nullable.hpp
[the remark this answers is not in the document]
JohanMabille: >  Should we just specify that if has_value() == false then that accessor provides a valid object with unspecified value? 

Yes, that's excatly it. I'll add more documentation to clarify things.

include/sparrow/nullable.hpp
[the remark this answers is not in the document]
JohanMabille: when moving / copying a nullable, you want to copy or move its underlying value and flag depending on whether the parameter is an lvalue or an rvalue, and depending on its nature: is it a nullable holding value or a nullable holding references. The different overloads of `values` and `null_flag` return the correct type in every situation.

Besides, constructor can indeed access the private members of another instance of the same type. But we also use these methods for conversion constructors and probably for efficient computation (when the number of null values in an array is low for instance).

include/sparrow/mp_utils.hpp
```diff
@@ -355,6 +355,15 @@ namespace sparrow::mpl
...
     //// Miscellaneous ///////////////////////////////
 
+    template <class T>
+    struct add_const_lvalue_reference
+        : std::add_lvalue_reference<std::add_const_t<T>>
+    {
+    };
+
+    template <class T>
+    using add_const_lvalue_reference_t = typename add_const_lvalue_reference<T>::type;
```
Alex-PLACET, line 365: Can you add unit tests ?

include/sparrow/mp_utils.hpp
```diff
@@ -410,4 +419,8 @@ namespace sparrow::mpl
 #endif
     }
 
+    /// The boolean_like concept specifies that a type can be convertible to and assignable from
+    /// bool.
+    template <class T>
+    concept boolean_like = std::constructible_from<T, bool> and std::convertible_to<T, bool>;
```
Alex-PLACET, line 425: Can you add unit tests ?

include/sparrow/nullable.hpp
Alex-PLACET, line 1: Formatting

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,649 @@
...
+            std::is_assignable_v<std::add_lvalue_reference_t<To2>, mpl::add_const_lvalue_reference_t<From2>>;
+
+        template <class To1, class From1, class To2, class From2>
+        concept both_assignable_from_cond_ref =
+            std::is_assignable_v<std::add_lvalue_reference_t<To1>, conditional_ref_t<To1, From1>> and
+            std::is_assignable_v<std::add_lvalue_reference_t<To2>, conditional_ref_t<To2, From2>>;
+
+        template <class T>
+        static constexpr bool is_nullable_v = mpl::is_type_instance_of_v<T, nullable>;
+    }
```
Alex-PLACET, line 171: Should we add test for these concepts ?

JohanMabille: I don't think so, they are indirectly tested trough all the constructor unit tests.

include/sparrow/nullable.hpp
JohanMabille: Let's do a final formatting pass and adjust the formatting rules, I am quite happy with this file actually.

Alex-PLACET (approved):

test/test_mpl.cpp
```diff
@@ -117,4 +117,24 @@ namespace sparrow
...
+    static_assert(not std::same_as<mpl::add_const_lvalue_reference_t<int&>, const int&>);
+
+    // boolean like
+    static_assert(mpl::boolean_like<bool>);
+
+    class like_a_bool
+    {
+    public:
+        like_a_bool& operator=(const bool&) { return *this; }
+        operator bool() const { return true; }
```
Klaim, line 136:
```suggestion
        explicit operator bool() const { return true; }
```

All well-behaved testable types has their bool operator explict.

test/test_nullable.cpp
```diff
@@ -12,55 +12,199 @@
...
 {
-    using nullable_double = nullable<double, bool>;
-    using nullable_int = nullable<int, bool>;
+    class Custom
+    {
+    public:
+
+        static int counter;
+
+        explicit Custom(const int& i = 0)
```
Klaim, line 30: Just `int i = 0` would have been enough 🤡

test/test_nullable.cpp
```diff
@@ -12,55 +12,199 @@
...
+
 #include "sparrow/nullable.hpp"
 
 #include "doctest/doctest.h"
 
 namespace sparrow
 {
-    using nullable_double = nullable<double, bool>;
-    using nullable_int = nullable<int, bool>;
+    class Custom
```
Klaim, line 24: Could you add a comment clarifying the intended properties of this type? So that we can check it matches the impl.

include/sparrow/mp_utils.hpp
```diff
@@ -410,4 +419,8 @@ namespace sparrow::mpl
 #endif
     }
 
+    /// The boolean_like concept specifies that a type can be convertible to and assignable from
```
Klaim, line 422:
```suggestion
    /// Matches types that can be convertible to and assignable from T
```

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,659 @@
...
+        bad_nullable_access& operator=(const bad_nullable_access&) noexcept = default;
+
+        virtual const char* what() const noexcept
+        {
+            return message;
+        }
+
+    private:
+
+        static constexpr const char* message = "Invalid access to null value";
```
Klaim, line 85:
```suggestion
        static constexpr const char* message = "Invalid access to null arrow value";
```

This idea of null value (which does not destroy the object) is specific to arrow and should be clear in the error, otherwise one would look for null pointers etc.

JohanMabille: It's more specific to libraries where you can have missing / null values. We could imagine moving this class to a util library later, taht could be used by data frame libraries / applications.

Klaim (approved):
We need experience with that type, the fact that `nullable<const T&, const bool&>` cannot be stored in a container makes it only useful as a return value of access functions; but that should be sufficient to start experimenting with.

include/sparrow/nullable.hpp
```diff
@@ -0,0 +1,662 @@
...
+#include <compare>
+#include <concepts>
+#include <exception>
+#include <type_traits>
+
+#include "sparrow/mp_utils.hpp"
+
+namespace sparrow
+{
+    using mpl::boolean_like;
```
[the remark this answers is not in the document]
JohanMabille: Good catch, will change that.

PR #136, opened 2024-07-09
Replace reference proxy with nullable

JohanMabille:
Depends on #135


include/sparrow/dictionary_encoded_layout.hpp
```diff
@@ -22,41 +22,66 @@
...
-     * @tparam SL the layout type of the dictionary.
-     * @tparam is_const a boolean flag specifying whether this iterator is const.
-     */
-    template <class IL, class SL, bool is_const>
-    class dictionary_value_iterator : public iterator_base<
-                                          dictionary_value_iterator<IL, SL, is_const>,
-                                          typename SL::value_type,
-                                          std::random_access_iterator_tag,
-                                          typename SL::const_reference>
+    template <class L, bool IC>
```
Klaim, line 25: Could you re-add the documentation?

include/sparrow/dictionary_encoded_layout.hpp
```diff
@@ -22,41 +22,66 @@
...
+    {
+        using layout_type = L;
+        using value_type = typename L::bitmap_value_type;
+        using tag = std::random_access_iterator_tag;
+        using const_reference = typename L::bitmap_const_reference;
+        static constexpr bool is_value = false;
+        static constexpr bool is_const = IC;
+    };
+
+    template <class Traits>
```
Klaim, line 47: Constraining the traits parameter would be helpful (something like `requires is_instance_of_v<dictionary_bitmap_traits>` would be enough I guess?)

include/sparrow/nullable.hpp
```diff
@@ -21,6 +21,16 @@
 
 #include "sparrow/mp_utils.hpp"
 
+// clang workaround: clang instantiates the constructor in SFINAE context,
+// which is incompatible with the implementation of standard libraries which
+// are not libc++.This leads to wrong compilation errors. Making the constructor
+// not constexpr prevents the compiler from instantiating it.
+#if defined(__clang__) && not defined(_LIBCPP_VERSION)
+#   define CONSTEXPR
```
Klaim, line 29: 1. Please prefix this define to avoid name collisions.
2. Is this a general issue? if yes, move this somewhere more general. 
3. If this means issues with `constexpr` will appear somewhere else, maybe we need an issue about this and have a pass where we replace `constexpr` by a define.

test/test_typed_array.cpp
```diff
@@ -360,18 +360,18 @@ TEST_SUITE("typed_array")
 
     TEST_CASE_TEMPLATE_INVOKE(
         all,
-        bool,
+        /*bool,
```
Klaim, line 363: I suppose that was temporary? The current build hasnt been tested yet will all the types below then.

include/sparrow/variable_size_binary_layout.hpp
```diff
@@ -109,9 +109,9 @@ namespace sparrow
         using inner_const_reference = CR;
         using bitmap_type = array_data::bitmap_type;
         using bitmap_const_reference = typename bitmap_type::const_reference;
-        using value_type = std::optional<inner_value_type>;
-        using reference = const_reference_proxy<self_type>;
-        using const_reference = const_reference_proxy<self_type>;
+        using value_type = nullable<inner_value_type>;
```
Alex-PLACET, line 112: include missing ?

include/sparrow/nullable.hpp
JohanMabille: 1. Sure, will do
2. Yes and no. I mean, the clang issue seems general, but it happens only with "incompatible" implementations of SFINAE.
3. So I don't know if we want to be conservative and use this macro only where we encounter the bug, or if we want to be extremely cautious and remove the constexpr everywhere when we build with clang and another standard library than libcxx.

include/sparrow/dictionary_encoded_layout.hpp
JohanMabille: It's either bitmap or value traits

SylvainCorlay (approved):

Alex-PLACET (approved):

PR #144, opened 2024-07-19
Implemented non const API of arrays

JohanMabille:


test/test_array.cpp
```diff
@@ -196,6 +196,30 @@ TEST_SUITE("array")
...
+            const auto& car = ar;
+            for (std::size_t i = 0; i < ar.size(); ++i)
+            {
+                if constexpr (std::same_as<T , bool>)
+                {
+                    std::get<ref>(ar.at(i)).value() = false;
+                    CHECK_EQ(std::get<const_ref>(car.at(i)).value(), false);
+                }
+                else if constexpr (std::same_as<T, std::string>)
+                {
```
Alex-PLACET, line 213: Why not implementing this part ?

JohanMabille: Good catch, that was a leftover

Klaim (approved):

PR #151, opened 2024-07-19
Array data to arrow array converter

Alex-PLACET:


include/sparrow/arrow_interface/array_data_to_arrow_array_converters.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+namespace sparrow
+{
+    /**
+     * Convert an array_data vector to an ArrowArray vector.
+     *
+     * @param ads The array_data vector to convert.
+     * @return The converted children.
+     */
+    std::vector<arrow_array_shared_ptr> children_from_array_data_vec(std::vector<array_data>&& ads);
```
Klaim, line 32: Why not take any range of `array_data`?

Klaim, line 32: I'm  not a fan of the naming of these functions, it's kind of redundant.
Taking this specific function as an example, maybe `to_arrow_array_vec(whatever_range_of_array_data)` would be better? I mean at reading site the "from" side of these names isnt really helpful.
Also it's best if the names reflect the general idea instead of the specifics, so that variations in arguments can have the same name.

Also, why is "children" mentioned here? The function is said to just be a conversion isnt it?

include/sparrow/arrow_interface/array_data_to_arrow_array_converters.hpp
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Convert an array_data to an ArrowArray.
+     * 
+     * @tparam T A const reference or rvalue reference to an array_data.
+     * @param ad The array_data to convert.
+     * @return The converted ArrowArray. 
+     */
+    template <class T>
+        requires std::same_as<std::remove_cvref_t<T>, array_data>
+    arrow_array_unique_ptr from_array_data(T&& ad);
```
Klaim, line 55: naming: go the other way around: `to_output_type(input)`
same logic for all such conversion functions

include/sparrow/arrow_interface/array_data_to_arrow_array_converters.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+    template <class T>
+        requires std::same_as<std::remove_cvref_t<T>, array_data>
+    std::vector<sparrow::buffer<uint8_t>> arrow_array_buffer_from_array_data(T&& ad)
+    {
+        std::vector<sparrow::buffer<uint8_t>> buffers;
+        buffers.reserve(ad.buffers.size() + 1);
+        buffers.emplace_back(std::move(ad.bitmap.buffer()));
+        if constexpr (std::is_rvalue_reference_v<T&&>)
```
Klaim, line 64: Shouldnt it be safer/clearer to check `decltype(ad)`?

include/sparrow/arrow_interface/array_data_to_arrow_array_converters.hpp
```diff
@@ -0,0 +1,123 @@
...
+        {
+            std::ranges::move(ad.buffers, std::back_inserter(buffers));
+        }
+        else
+        {
+            std::ranges::copy(ad.buffers, std::back_inserter(buffers));
+        }
+        return buffers;
```
Klaim, line 72: how is this implementation different from `return std::vector<sparrow::buffer<uint8_t>>(ad.buffers.begin(), ad.buffers.end())`?

include/sparrow/arrow_interface/array_data_to_arrow_array_converters.hpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    template <class T>
+        requires std::same_as<std::remove_cvref_t<T>, array_data>
+    inline arrow_array_unique_ptr from_array_data(T&& ad)
```
Klaim, line 77: `inline` is already implied by `template` and `constexpr` for functions.

DerThorsten (approved):

include/sparrow/arrow_interface/array_data_to_arrow_array_converters.hpp
Alex-PLACET: The first element of "buffers" is the bitmap buffer

Klaim (approved):

PR #152, opened 2024-07-22
Added Data Storage template parmeter to layout classes

### LINKED ISSUE #139: Layout template parameter for the storage
The layout classes should have an additional template parameter for specifying the storage. Currently, they work with `array_data` only, which makes it complicated (if not impossible) to transfer data from [ArrowArray structures](https://arrow.apache.org/docs/dev/format/CDataInterface.html#structure-definitions).

The layout classe should be able to work with these structures. the [buffer_view](https://github.com/man-group/sparrow/blob/main/include/sparrow/buffer_view.hpp) and [dynamic_bitset_view](https://github.com/man-group/sparrow/blob/main/include/sparrow/dynamic_bitset.hpp#L166) classes can be built on top of the C strutures to ease their manipulation.

Free functions for retrieving the buffers / bitmap / children members should be added to avoid specializing the layout for each storage type (currently `array_data` and `ArrowArray`, but there might be others in the future).

JohanMabille:
Fixes #139

### DIFF
--- a/include/sparrow/array/array_data.hpp
+++ b/include/sparrow/array/array_data.hpp
@@ -80,9 +80,9 @@ namespace sparrow
         { type_descriptor(t) } -> std::same_as<data_descriptor>;
         { length(t) } -> std::same_as<std::int64_t>;
         { offset(t) } -> std::same_as<std::int64_t>;
-        bitmap(t);
+        { bitmap(t) } -> std::ranges::random_access_range;
         { buffers_size(t) } -> std::same_as<std::size_t>;
-        buffer_at(t, i);
+        { buffer_at(t, i) } -> std::ranges::random_access_range;
         { child_data_size(t) } -> std::same_as<std::size_t>;
         child_data_at(t, i);
         dictionary(t);
--- a/include/sparrow/array/array_data.hpp
+++ b/include/sparrow/array/array_data.hpp
@@ -52,6 +52,38 @@ namespace sparrow
         value_ptr<array_data> dictionary;
     };
 
+    data_descriptor type_descriptor(const array_data& data);
+    array_data::length_type length(const array_data& data);
+    std::int64_t offset(const array_data& data);
+
+    array_data::bitmap_type& bitmap(array_data& data);
+    const array_data::bitmap_type& bitmap(const array_data& data);
+
+    std::vector<array_data::buffer_type>& buffers(array_data& data);
+    const std::vector<array_data::buffer_type>& buffers(const array_data& data);
+
+    std::vector<array_data>& child_data(array_data& data);
+    const std::vector<array_data>& child_data(const array_data& data);
+
+    value_ptr<array_data>& dictionary(array_data& data);
+    const value_ptr<array_data>& dictionary(const array_data& data);
+
+    /**
+     * Concept for a structure that can be used as a data storage in the layout and the
+     * typed_array class.
+     */
+    template <class T>
+    concept data_storage = requires(T t)
+    {
+        type_descriptor(t);
+        length(t);
+        offset(t);
+        bitmap(t);
+        buffers(t);
+        child_data(t);
+        dictionary(t);
+    };
+
     /**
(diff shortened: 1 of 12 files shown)


include/sparrow/array/array_data.hpp
```diff
@@ -52,6 +52,42 @@ namespace sparrow
...
+    const value_ptr<array_data>& dictionary(const array_data& data);
+
+    /**
+     * Concept for a structure that can be used as a data storage in the layout and the
+     * typed_array class.
+     */
+    template <class T>
+    concept data_storage = requires(T t, std::size_t i)
+    {
+        type_descriptor(t);
```
Klaim, line 80: I think we can add `-> expected-return-type` on every function call here? Maybe for some the return type can just be a concept.

include/sparrow/array/array_data.hpp
```diff
@@ -52,6 +52,42 @@ namespace sparrow
...
+
+    value_ptr<array_data>& dictionary(array_data& data);
+    const value_ptr<array_data>& dictionary(const array_data& data);
+
+    /**
+     * Concept for a structure that can be used as a data storage in the layout and the
+     * typed_array class.
+     */
+    template <class T>
+    concept data_storage = requires(T t, std::size_t i)
```
Klaim, line 78: Could you add tests with a custom `data_storage` (designed for testing) that just checks that it is compiling with that concept and used correctly in layouts?

include/sparrow/array/array_data.hpp
```diff
@@ -99,6 +135,75 @@ namespace sparrow
...
+    }
+
+    inline std::size_t buffers_size(const array_data& data)
+    {
+        return data.buffers.size();
+    }
+
+    inline array_data::buffer_type& buffer_at(array_data& data, std::size_t i)
+    {
+        return data.buffers[i];
```
Klaim, line 174: We might want to add assertions in accessors which take a size, that will be helpful.

include/sparrow/array/array_data.hpp
JohanMabille: For most of the functions, yes. For the buffer and bitmap accessors, we cannot really because the return types are really different.

Klaim: Can at least a concept be specified? if not nevermind

JohanMabille: We can require a random_access_range, it's not complete but at least covers most of our usages.

include/sparrow/array/array_data.hpp
```diff
@@ -52,6 +52,42 @@ namespace sparrow
...
+    concept data_storage = requires(const T t, std::size_t i)
+    {
+        { type_descriptor(t) } -> std::same_as<data_descriptor>;
+        { length(t) } -> std::same_as<std::int64_t>;
+        { offset(t) } -> std::same_as<std::int64_t>;
+        { bitmap(t) } -> std::ranges::random_access_range;
+        { buffers_size(t) } -> std::same_as<std::size_t>;
+        { buffer_at(t, i) } -> std::ranges::random_access_range;
+        { child_data_size(t) } -> std::same_as<std::size_t>;
+        child_data_at(t, i);
```
Klaim, line 87:
```suggestion
        { child_data_at(t, i); } -> std::integer;
```
Would that be reasonable?

include/sparrow/array/array_data.hpp
```diff
@@ -52,6 +52,42 @@ namespace sparrow
...
+        dictionary(t);
```
Klaim, line 88: Can we require some specific interface that looks like value_ptr?

Klaim (approved):
LGTM as is, potential improvements suggested.

include/sparrow/array/array_data.hpp
JohanMabille: actually `child_data_at` returns something that must satisfy the `data_storage` concept, so it's kind of recursive.

PR #153, opened 2024-07-22
Layout propagation

### LINKED ISSUE #140: typed_array template parameter for the storage
When the layout classes support additional storage types (see #139 ), this should be propagated to the `typed_array` class, since it is responsible for holding the storage.

JohanMabille:
Fix #140 
Based on #152


Klaim (approved):
Conceptually LGTM although it needs a rebase. If there is a need to cahnge something here after the rebase, then ping me again to re-review.

Klaim (approved):

PR #154, opened 2024-07-23
Implemented missing dynamic_bitset_view constructor

JohanMabille:


test/test_dynamic_bitset.cpp
```diff
@@ -23,6 +23,8 @@ namespace sparrow
 {
     struct bitmap_fixture
     {
+        using buffer_type = std::uint8_t*;
+
```
Klaim, line 27: Could you also delete copy and move operations to avoid multiple deletes if the object is accidentally moved/copied?

Klaim (approved):

PR #163, opened 2024-08-05
Generic traits for standard base types (partial fix #162)

### LINKED ISSUE #162: Incorrect assumption that fixed floating types will always be aliases of `double` and `float` leads to errors
Reported through `build2`'s ci (using my wip packaging) : 
- ci output: https://ci.stage.build2.org/@6cbc0f07-2e1b-43ce-a5f5-5c4e37f2ac92
- example of error report: https://ci.stage.build2.org/@6cbc0f07-2e1b-43ce-a5f5-5c4e37f2ac92/sparrow/0.0.3-a.0.20240726155934.0f12fb036311/log/x86_64-linux-gnu/linux_debian_12-gcc_14-ndebug_O3/default/stage/0.18.0-a.0.20240628122722.68eac2eacdac/test

In the configurations we build and test on github, `float32_t` and `float64_t` usually are aliases to `double` and `float`. However in some other configurations we didnt test, these names are not aliases, they are concrete separate types (note also the possibility coming from using our float16_t implementation depending on if the standard library available provides or not fixed floating point types). In our tests, we use `double` directly in some places where we end up with `arrow_traits<double>`. When `float64_t` is an alias to `double` (like on our ci), `arrow_traits<double>` and `arrow_traits<float64_t>` are the same type. When it's not (some configurations on `build2`'s CI) these are different types.
At the moment we only define `arrow_traits<float64_t>`, therefore `arrow_traits<double>` is never found at compile-time and support for `double` is therefore non-existent in these configurations, leading to compilation errors when trying to use `double` in sparrow.

Attempt to fix that issue by adding `arrow_traits<double>` (and `float`) leads to duplicate type definition errors in configurations where `float64_t` are aliases to `double`, so it's the reverse problem.

In my experiments, the best way to fix this seems to be to have only one `arrow_traits<std::floating_point>` template which handles all the possible floating point types standard at least. In that case, `type_id` needs to be set to a value depending on the size of the type. My current experiment half-works (it reveals other similar issues) and I intend to complete it to completely solve the issue.

Note that the same solution can be set for integers, I'll try to see if that works too. That would prevent hitting similar issues on exotic platforms with integers and actually reduce the code size for arrow traits provided by default.

Klaim:
Rewrite the `arrow_traits` specializations for standard integral and floating-point types support.

The main goal is to fix #162 by providing generic traits that will adapt to both situations:
- when `double` and `float64_t` are aliases (1 traits specialization generated for the real type)
- when they are not aliases but distinct types (2 traits spetialization generated, one for each distinct type)

Although the issue was originally spotted with `double` usage, I generalized to all floating-point and integer types as the issue might arise with these types too. It also reduces the code a lot.

### DIFF
--- a/include/sparrow/array/data_traits.hpp
+++ b/include/sparrow/array/data_traits.hpp
@@ -47,59 +47,14 @@ namespace sparrow
         static constexpr data_type type_id = data_type::BOOL;
     };
 
-    template <>
-    struct arrow_traits<std::uint8_t> : common_native_types_traits<std::uint8_t>
+    // Define automatically all standard integral types support (including `bool`).
+    template <class T>
+        requires std::integral<T>
+    struct arrow_traits<T> : common_native_types_traits<T>
     {
-        static constexpr data_type type_id = data_type::UINT8;
+        static constexpr data_type type_id = data_type_from_size<T>();
     };
 
-    template <>
-    struct arrow_traits<std::int8_t> : common_native_types_traits<std::int8_t>
-    {
-        static constexpr data_type type_id = data_type::INT8;
-    };
-
-    template <>
-    struct arrow_traits<char> : common_native_types_traits<char>
-    {
-        static constexpr data_type type_id = data_type::UINT8;
-    };
-
-    template <>
-    struct arrow_traits<std::uint16_t> : common_native_types_traits<std::uint16_t>
-    {
-        static constexpr data_type type_id = data_type::UINT16;
-    };
-
-    template <>
-    struct arrow_traits<std::int16_t> : common_native_types_traits<std::int16_t>
-    {
-        static constexpr data_type type_id = data_type::INT16;
-    };
-
-    template <>
-    struct arrow_traits<std::uint32_t> : common_native_types_traits<std::uint32_t>
-    {
-        static constexpr data_type type_id = data_type::UINT32;
-    };
-
-    template <>
-    struct arrow_traits<std::int32_t> : common_native_types_traits<std::int32_t>
-    {
-        static constexpr data_type type_id = data_type::INT32;
-    };
-
-    template <>
-    struct arrow_traits<std::uint64_t> : common_native_types_traits<std::uint64_t>
-    {
-        static constexpr data_type type_id = data_type::UINT64;
-    };
-
-    template <>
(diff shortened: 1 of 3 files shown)


Klaim:
Note: this is still a draft because [build2's CI still point to other related errors on some platforms we dont test here](https://ci.stage.build2.org/@31e628c7-67d7-4179-8e14-807e2f196006).

Klaim:
@SylvainCorlay 
> Are you trying to get all build2 tests to pass on this PR or only those relared to float aliases?

At the moment I focus on the issues triggered by the alias-or-not-alias problem yes. `build2`'s failing ci builds only report issues related to that at the moment, or did I miss another issue?

Klaim:
@SylvainCorlay 
Also some notes: https://github.com/man-group/sparrow/issues/162#issuecomment-2269363691

I think this PR only partially fixes the issue, I can't find a proper solution for the other problemes the alias-or-not problem causes.
I guess we can review this at least, it does solve part of the issue but it's not complete to remove it.

Klaim:
Let's review this first while I try to find a solution for the other blocking issues.

Klaim:
I'll add some separate tests for the new functions.

include/sparrow/array/data_traits.hpp
```diff
@@ -39,83 +41,23 @@ namespace sparrow
...
-    template <>
-    struct arrow_traits<float32_t> : common_native_types_traits<float32_t>
+    // Define automatically all standard floating-point types support.
+    template<class T>
+        requires std::floating_point<T>
+    struct arrow_traits<T> : common_native_types_traits<T>
     {
-        static constexpr data_type type_id = data_type::FLOAT;
+        static constexpr data_type type_id = data_type_from_size<T>();
     };
```
Alex-PLACET, line 59: Can't you merge with the previous definition by merging the requires or creating an "arithmetic" concept ?

Klaim: You mean using `std::integer<T> or std::floating_point<T>` ? Yes that would make sense, although I'll refrene from making an arithmetic concept for now because I suspect it implies additional expectations (not sure) but I can at least merge both specialisations with these constraintes yes, will do.

Alex-PLACET (approved):

JohanMabille (approved):

PR #164, opened 2024-08-06
factory function refactor

DerThorsten:
* use pre-existing factory functions
* use values for bitmaps instead of const reference
* remove requirement for const_range (does not work for std::vector<bool> because our concept check for const_ranges is not working for std::vector<bool> )
* added `is_mutable` to array_data / external_array_data

The constructor from ranges for ranges of nullables will be added in a followup PR

### DIFF
--- a/include/sparrow/array/array_data_factory.hpp
+++ b/include/sparrow/array/array_data_factory.hpp
@@ -14,8 +14,6 @@
 
 #pragma once
 
-#include <iostream> // REMOVE ME
-
 #include <ranges>
 #include <algorithm>
 #include <cstddef>
@@ -140,18 +138,14 @@ namespace sparrow
         // Check that the range is a range of ranges
         if constexpr (std::ranges::range<T>)
         {
-            std::cout<<"is range"<<std::endl;
             SPARROW_ASSERT_TRUE(check_all_elements_have_same_size(values));
         }
 
-        std::cout<<"values.size(): "<<values.size()<<std::endl;
-        std::cout<<"bitmap.size(): "<<bitmap.size()<<std::endl;
         SPARROW_ASSERT_TRUE(values.size() == bitmap.size());
         SPARROW_ASSERT_TRUE(std::cmp_greater_equal(values.size(), offset));
 
         const auto create_buffer = [&values]()
         {
-            std::cout<<"create_buffer"<<std::endl;
             const size_t buffer_size = (values.size() * sizeof(T)) / sizeof(uint8_t);
             array_data::buffer_type buffer(buffer_size);
             auto data = buffer.data<T>();
@@ -160,7 +154,6 @@ namespace sparrow
                 *data = value;
                 ++data;
             }
-            std::cout<<"buffer_size: "<<buffer_size<<std::endl;
             return buffer;
         };
         using U = std::conditional_t<std::same_as<T, std::string_view>, std::string, T>;
@@ -211,7 +204,7 @@ namespace sparrow
     template <range_for_array_data ValueRange>
     array_data make_array_data_for_variable_size_binary_layout(
         ValueRange&& values,
-        const array_data::bitmap_type& bitmap,
+        array_data::bitmap_type bitmap,
         std::int64_t offset
     )
     {
@@ -264,12 +257,11 @@ namespace sparrow
 
         using T = std::unwrap_ref_decay_t<std::unwrap_ref_decay_t<std::ranges::range_value_t<ValueRange>>>;
         using U = get_corresponding_arrow_type_t<T>;
-        std::cout<<"return..."<<std::endl;
         return {
             .type = data_descriptor(arrow_type_id<U>()),
             .length = static_cast<array_data::length_type>(values.size()),
             .offset = offset,
-            .bitmap = bitmap,
+            .bitmap = std::move(bitmap),
             .buffers = create_buffers(),
             .child_data = {},
             .dictionary = nullptr
@@ -387,7 +379,7 @@ namespace sparrow
     template <range_for_array_data ValueRange>
     array_data make_array_data_for_dictionary_encoded_layout(
         ValueRange&& values,
-        const array_data::bitmap_type& bitmap,
+        array_data::bitmap_type bitmap,
         std::int64_t offset
     )
     {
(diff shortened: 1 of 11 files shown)


include/sparrow/array/array_data_factory.hpp
```diff
@@ -487,21 +489,45 @@ namespace sparrow
...
     template <arrow_layout Layout, std::ranges::input_range ValueRange>
-        requires(!mpl::constant_range<ValueRange>)
     array_data
-    make_default_array_data(ValueRange&& values, const array_data::bitmap_type& bitmap, std::int64_t offset)
+    make_default_array_data(ValueRange&& values)
+    {
+        const std::size_t size = std::ranges::size(values);
+        array_data::bitmap_type bitmap(size, true);
+        const std::int64_t offset = 0;
+        return make_default_array_data<Layout>(std::forward<ValueRange>(values), std::move(bitmap), offset);
```
Alex-PLACET, line 508: Could be all constexpr if we change dynamic_bitset to constexpr too. Maybe add a todo.

include/sparrow/array/external_array_data.hpp
```diff
@@ -96,6 +96,7 @@ namespace sparrow
     {
     public:
 
+        static constexpr bool is_mutable = false;
```
Klaim, line 99: To complete this addition, it would be best to also add a requirement in the [`data_storage`](https://github.com/man-group/sparrow/blob/main/include/sparrow/array/array_data_concepts.hpp#L59-L75) for this member to exist and be convertible to bool, that way we're sure future breaking changes will get caught even before using these types in layout impls.

include/sparrow/array/typed_array.hpp
```diff
@@ -31,6 +31,28 @@
...
+    {
+    };
+
+    template <class T, class L>
+    struct is_typed_array_impl<typed_array_impl<T, L>> : std::true_type
+    {
+    };
+
+    template <class A>
+    constexpr bool is_typed_array_impl_v = is_typed_array_impl<A>::value;
```
Klaim, line 53: I suspect this could be implemented using `mpl::is_type_instance_of/_v` ([see there](https://github.com/man-group/sparrow/blob/main/include/sparrow/utils/mp_utils.hpp#L32-L46)) as this is essentially `is_type_instance_of_v<typed_array_impl<T,L>>` ?
I see it's just existing code moved here so not necessarilly tweaked in this pr, as you prefer.

include/sparrow/array/typed_array.hpp
```diff
@@ -75,15 +97,26 @@ namespace sparrow
         typed_array_impl& operator=(typed_array_impl&& rhs);
         ///@}
 
+        // /** Construct a typed array from a range of values.
```
Klaim, line 100: Something went wrong with the comments here 😆

include/sparrow/array/array_data_factory.hpp
```diff
@@ -487,21 +489,45 @@ namespace sparrow
...
+    requires is_arrow_base_type_extended<std::decay_t<T>>
+    array_data make_default_array_data( 
+        typename Layout::size_type n
+        ,  T && value)
     {
-        return make_default_array_data<Layout>(std::as_const(values), bitmap, offset);
+        auto repeated_range = std::ranges::iota_view{size_t(0),size_t(n)} | std::views::transform([&](auto) { return value; });
+
+        return make_default_array_data<Layout>(repeated_range);
     }
```
Alex-PLACET, line 530: All constexpr

include/sparrow/array/array_data_factory.hpp
```diff
@@ -126,7 +127,7 @@ namespace sparrow
      * @param offset The offset of the array data.
      * @return The created array_data object.
      */
-    template <constant_range_for_array_data ValueRange>
+    template <range_for_array_data ValueRange>
```
Klaim, line 130: Just a note: looking at this concept and it's constant version and trying to understand what's wrong lead me to think `constant_range` and `constant_iterator` are not correct. If you are not already on it I'll take note to investigate this.

test/test_typed_array.cpp
```diff
@@ -520,20 +513,4 @@ TEST_SUITE("typed_array")
         float64_t
     );
 
-    TEST_CASE_TEMPLATE_INVOKE(
```
Klaim, line 523: Just to clarify, this was removed because it's duplicating the previous block?

Klaim (approved):
LGTM overall, pointed a few improvements

include/sparrow/array/array_data_factory.hpp
DerThorsten: I did not yet tried to fix it, but they indeed are not working as expected, at least for std::vector<bool>

But I believe the restriction to a const_range is un-needed anyhow

test/test_typed_array.cpp
Alex-PLACET: The difference isthe std::string entry

DerThorsten: We used to run the "constructor from n-times same value" for all types **except string**
I removed this limitation, therefore we need not extra `TEST_CASE_TEMPLATE_INVOKE` and can just put the "constructor from n-times same value" in the pre-existing invocation

include/sparrow/array/array_data_factory.hpp
DerThorsten: are you suggesting to make the whole `make_default_array_data` function and everything this function calls constexpr?

include/sparrow/array/array_data_factory.hpp
DerThorsten: are you suggesting to make the whole `make_default_array_data` function and everything this function calls constexpr?

include/sparrow/array/typed_array.hpp
DerThorsten: I just moved it up because its needed in the requirements for the "from-range" constructor.
We need to make sure the range-constructor is not called for typed-arrays themself

include/sparrow/array/array_data_factory.hpp
Klaim: My suspicion is that these concepts should be requiring that `std::ranges::cbegin/cend` works, instead of `std::ranges::begin/end` although I didnt test this theory yet.

include/sparrow/array/array_data_factory.hpp
Alex-PLACET: yes but it requires to change dynamic_bitset class which is not the point of this PR. So maybe just write a TODO comment and we will do that later.

Alex-PLACET (approved):

include/sparrow/array/array_data_factory.hpp
DerThorsten: In that case I prefer to add an new issue instead of TODOs in the code

Klaim (approved):

include/sparrow/array/array_data_factory.hpp
DerThorsten: https://github.com/man-group/sparrow/issues/166

PR #165, opened 2024-08-06
Buffer adaptor with immutable input

Alex-PLACET:
buffer_adaptor can now takes as input a const reference to a container. In that case all the const methods are disabled.

### DIFF
--- a/include/sparrow/buffer/buffer_adaptor.hpp
+++ b/include/sparrow/buffer/buffer_adaptor.hpp
@@ -15,19 +15,29 @@
 #pragma once
 
 #include <cmath>
-#include <limits>
+#include <functional>
+#include <ranges>
+#include <type_traits>
 #include <utility>
 
-#include "sparrow/buffer/buffer.hpp"
 #include "sparrow/utils/contracts.hpp"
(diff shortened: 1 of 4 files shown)


include/sparrow/buffer/buffer_adaptor.hpp
```diff
@@ -15,19 +15,33 @@
...
+
 
 namespace sparrow
 {
+    template <typename FromBufferRef, typename T>
+    concept BufferReference = std::ranges::contiguous_range<FromBufferRef> && std::is_reference_v<FromBufferRef>
+                              && (sizeof(std::ranges::range_value_t<FromBufferRef>) <= sizeof(T));
+
+    template <typename FromBufferRef, typename T>
+    concept T_is_const_if_FromBufferRef_is_const = mpl::T_matches_qualifier_if_Y_is<T, FromBufferRef, std::is_const>;
```
DerThorsten, line 33: this could really need a better name

DerThorsten (approved):

SylvainCorlay:
> In that case all the const methods are disabled.

You meant the non-const :)

PR #169, opened 2024-08-07
Nullable range constructor

DerThorsten:
* use a range of bools instead of a bitmap in the factory functions:
    * this allows to split a vector<nullable<T>> into two ranges, the values and the bitmaps
    * this can then be passed to pre-existing factory functions
    * note that this changes how the length of missing values is accounted for in the dynamic layout (see below)   

This changes some behavior:
```
  values = ["fo", "bar", "fubar"]
  bitmap  = [1,0,1]  
```

used to account space  for "bar" in the offsets.
This PR changes this st. we will **not take** the length of "bar" into account since its a missing value (ie the effective length will be 0)

Why? This is the only consistent way of doing it:

Consider a range of nullables:
```C++
std::vector<nullable<std::string>> {
   nullable("fo"), nullval, nullval("fubar")
}
```
Here it's clear that we cannot give the missing value any length in the offsets, therefore we should also not account for space in the case above. 

* due to this behavior change the test for the dynamic layout had to be changed

### DIFF
--- a/include/sparrow/array/array_data_factory.hpp
+++ b/include/sparrow/array/array_data_factory.hpp
@@ -32,6 +32,7 @@
 #include "sparrow/layout/dictionary_encoded_layout.hpp"
 #include "sparrow/layout/fixed_size_layout.hpp"
 #include "sparrow/layout/variable_size_binary_layout.hpp"
+#include "sparrow/utils/nullable.hpp"
 #include "sparrow/utils/contracts.hpp"
 #include "sparrow/utils/memory.hpp"
 #include "sparrow/utils/mp_utils.hpp"
@@ -39,34 +40,12 @@
 
 namespace sparrow
 {
-
-    /*
-    * \brief Concept to check if a range is convertible to a range of booleans.
-    *
-    * A range is considered convertible to a range of booleans if it is a range and its value type is convertible to a boolean.
-    *
-    * @tparam BoolRange The range to check.
-    */
-    template<class BoolRange>
-    concept bool_convertible_range = std::ranges::range<BoolRange> &&
-        std::convertible_to<std::ranges::range_value_t<BoolRange>, bool>;
-
-    /*
-    * \brief Concept to check if a range is a range of nullables.
-    *
-    * A range is considered a range of nullables if it is a range and its value type is a nullable.
-    *
-    * @tparam RangeOfNullables The range to check.
-    */
-    template<class RangeOfNullables>
-    concept range_of_nullables = std::ranges::range<RangeOfNullables> && is_nullable<std::ranges::range_value_t<RangeOfNullables>>::value;
-
     /// \cond
     namespace detail
     {
 
         // a helper function to make the bitmap from a range
-        template<bool_convertible_range BoolRange>
+        template<mpl::bool_convertible_range BoolRange>
         requires (!std::is_same_v<std::decay_t<BoolRange>, array_data::bitmap_type>)
         array_data::bitmap_type make_array_data_bitmap(BoolRange&& range)
         {
@@ -182,7 +161,7 @@ namespace sparrow
      * @param offset The offset of the array data.
      * @return The created array_data object.
      */
-    template <range_for_array_data ValueRange, bool_convertible_range BitmapRange>
+    template <range_for_array_data ValueRange, mpl::bool_convertible_range BitmapRange>
     array_data make_array_data_for_fixed_size_layout(
         ValueRange&& values,
         BitmapRange && bitmap,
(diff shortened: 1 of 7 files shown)


include/sparrow/array/array_data_factory.hpp
```diff
@@ -39,6 +39,48 @@
...
+
+    namespace detail
+    {
+
+        // a helper function to make the bitmap from a range
+        template<bool_convertible_range BoolRange>
+        array_data::bitmap_type make_array_data_bitmap(BoolRange&& range)
+        {
+            array_data::bitmap_type bitmap(std::ranges::size(range), true);
+            for(size_t i = 0; auto value : range)
```
SylvainCorlay, line 63: std::size_t

SylvainCorlay: (at other places too)

DerThorsten: I noticed that the library used "naked" size_t in various places. Should we are it should be `std::size_t` everywhere (I would prefer that)

include/sparrow/array/array_data_factory.hpp
```diff
@@ -39,6 +39,48 @@
...
+            {
+                if(!value)
+                {
+                    bitmap[i] = false;
+                }
+                ++i;
+            }
```
Alex-PLACET:
```suggestion
            std::ranges::transform(range, bitmap.begin(), [](auto value) { return static_cast<bool>(value); });
```

include/sparrow/array/array_data_factory.hpp
```diff
@@ -39,6 +39,48 @@
...
+
+    // a range where the value type is convertible to bool
+    template<class BoolRange>
+    concept bool_convertible_range = std::ranges::range<BoolRange> &&
+        std::convertible_to<std::ranges::range_value_t<BoolRange>, bool>;
+
+    
+    // a range of nullable values / nullable<T>
+    template<class RangeOfNullables>
+    concept range_of_nullables = std::ranges::range<RangeOfNullables> && is_nullable<std::ranges::range_value_t<RangeOfNullables>>::value;
```
Alex-PLACET: Could be moved to mp_utils

include/sparrow/array/array_data_factory.hpp
SylvainCorlay: Yes. I noticed too. One weird thing is in one place, when I replaced with std::size_t, behavior changed! (warnings about signeness appeared).

In any case, std:: should be required.

include/sparrow/array/array_data_factory.hpp
```diff
@@ -140,30 +182,37 @@ namespace sparrow
...
-        const auto create_buffer = [&values]()
+        const auto size = std::ranges::size(values);
+        SPARROW_ASSERT_TRUE(size == std::ranges::size(bitmap));
+        SPARROW_ASSERT_TRUE(std::cmp_greater_equal(size, offset));
+
+
+        std::vector<array_data::buffer_type> buffers(1);
+        const size_t buffer_size = (size * sizeof(T)) / sizeof(uint8_t);
+        array_data::buffer_type buffer(buffer_size);
+        auto data = buffer.data<T>();
```
Alex-PLACET, line 190: Should be replaced by `buffer_adaptor`

DerThorsten: Fyi, I only tried to refactor this, such that it is not using a lambda function`create_buffers` any more.
At the time the lambda function `create_buffers` used to be called, the bitmap-range has already been moved into the array-data struct, therefore I needed to create the buffers before the assignment of `.bitmap = `.

Therefore it did not make sense anymore to do this in a lambda function

include/sparrow/array/array_data_factory.hpp
DerThorsten: not sure about that, that would make mpl utils depend on `nullable.hpp`.

We could move `range_of_nullables` to `nullable.hpp` and `bool_convertible_range` to mpl utils

include/sparrow/array/array_data_factory.hpp
DerThorsten: I would prefer to switch to the buffer_adaptor in a follow-up PR.

include/sparrow/array/array_data_factory.hpp
```diff
@@ -490,6 +537,46 @@ namespace sparrow
     }
 
 
+
+    /**
+     * \brief Creates a default array_data object with the specified layout and values and indicate which values are missing.
```
Klaim, line 542:
```suggestion
     * Creates a default array_data object with the specified layout and values and indicate which values are missing.
```
We use auto-brief :) same suggestion for other comments below

include/sparrow/utils/nullable.hpp
```diff
@@ -55,6 +55,17 @@ namespace sparrow
     template<class N, class T>
     concept is_nullable_of_convertible_to = is_nullable_v<N> &&  std::convertible_to<typename N::value_type, T>;
 
+    /*
+    * \brief Concept to check if a range is a range of nullables.
```
Klaim, line 59:
```suggestion
    * Matches a range of nullables objects.
```

Klaim (approved):
LGTM, only doc nitpicking 👍🏽

Alex-PLACET (approved):

PR #177, opened 2024-08-13
Add modifiers to dynamic_bitset, and add constexpr

Alex-PLACET:


include/sparrow/buffer/dynamic_bitset.hpp
```diff
@@ -388,45 +424,95 @@ namespace sparrow
...
+                "dynamic_bitset_base::at: index out of range for dynamic_bitset_base of size"
+                + std::to_string(size()) + " at index " + std::to_string(pos)
+            );
+        }
+        return test(pos);
+    }
+
+    template <random_access_range B>
+    constexpr auto dynamic_bitset_base<B>::front() -> reference
+    {
```
Klaim, line 492: Add assertions in non-throwing functions that access the elements (front, back, etc.)?

include/sparrow/buffer/dynamic_bitset.hpp
```diff
@@ -561,24 +647,156 @@ namespace sparrow
         zero_unused_bits();
     }
 
+    template <random_access_range B>
+    constexpr void dynamic_bitset_base<B>::clear() noexcept
+    {
+        resize(0);
```
Klaim, line 653: This calls buffer's resize which will call the destructor of the storage type. Just making sure I understand correctly: is this always equivalent to just setting every bytes to zero? (maybe we should constrain the storage type to only be a basic type?)

include/sparrow/buffer/dynamic_bitset.hpp
```diff
@@ -561,24 +647,156 @@ namespace sparrow
...
+        SPARROW_ASSERT_TRUE(cbegin() <= pos);
+        SPARROW_ASSERT_TRUE(pos <= cend());
+        const auto index = static_cast<size_type>(std::distance(cbegin(), pos));
+        const size_type old_size = size();
+        const size_type count = static_cast<size_type>(std::distance(first, last));
+        const size_type new_size = old_size + count;
+
+        resize(new_size);
+
+        // TODO: The current implementation is not efficient. It can be improved.
```
Klaim, line 705: Consider creating an issue about this type's performance once this is merged?

Klaim (approved):
LGTM, just needs more assertions 👍🏼

include/sparrow/buffer/dynamic_bitset.hpp
Alex-PLACET: Ok added: #201

include/sparrow/buffer/dynamic_bitset.hpp
Alex-PLACET: I modified the code and is now:
```
m_buffer.clear();
m_size = 0;
m_null_count = 0;
```

include/sparrow/buffer/dynamic_bitset.hpp
Alex-PLACET: done

JohanMabille (approved):

PR #205, opened 2024-09-09
Array hierarchy

JohanMabille:


include/sparrow_v01/layout/array_base.hpp
```diff
@@ -0,0 +1,121 @@
...
+        , m_bitmap(init_bitmap())
+    {
+    }
+
+    inline auto array_with_bitmap::bitmap() const -> const bitmap_type&
+    {
+        return m_bitmap;
+    }
+
+    inline auto array_with_bitmap::init_bitmap() const -> bitmap_type
```
Alex-PLACET, line 116: Not all layouts have a bitmap. We should check, and maybe the proxy can expose it, which kind of layout it is and returns an optional ?

include/sparrow_v01/layout/array_base.hpp
```diff
@@ -0,0 +1,121 @@
...
+    {
+        return bitmap_type(array_base::data().buffers()[0].data(),
```
Alex-PLACET, line 118: You should check the buffer size.

include/sparrow_v01/layout/array_base.hpp
Alex-PLACET, line 116: Except that the C structures are created from sparrow, we can't reallocate buffers. So we have to return a view.

JohanMabille: My idea was to have this information encoded at build time rather than at runtime:
- layouts with bitmap would inherit from `array_with_bitmap`
- layouts without bitmap (like the null_array) would niherit from `array_base`

JohanMabille: `bitmap_type` is defined to ba a view in [array_with_bitmap](https://github.com/man-group/sparrow/pull/205/files#diff-4efc2a1923bf89abb9e10afc95f3e32aea3f5af1cc0a026c0131959c8a991bddR58)

DerThorsten (changes requested):

include/sparrow_v01/layout/array_base.hpp
```diff
@@ -0,0 +1,121 @@
...
+    protected:
+
+        array_base(arrow_proxy proxy);
+        array_base(const array_base&) = default;
+
+        const arrow_proxy& data() const;
+
+    private:
+
+        virtual array_base* clone_impl() const = 0;
```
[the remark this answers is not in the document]
JohanMabille: I don't think the additional cost of the virtual method would be that much compared to the copy of the data. But that something we can change when we implement the visitor if we need it.

DerThorsten (approved):

test_v01/test_null_array.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include "doctest/doctest.h"
+
+#include "../test/external_array_data_creation.hpp"
+
+namespace sparrow
+{
+    arrow_proxy make_proxy_for_null_array(std::size_t size)
+    {
+        ArrowSchema sc;
```
Klaim, line 27:
```suggestion
        ArrowSchema sc{};
```
To silence the static analysis warnings

test_v01/test_null_array.cpp
```diff
@@ -0,0 +1,99 @@
...
+        ArrowArray ar;
```
Klaim, line 28:
```suggestion
        ArrowArray ar{};
```

test_v01/test_null_array.cpp
```diff
@@ -0,0 +1,99 @@
...
+        test::fill_schema_and_array<null_type>(sc, ar, size, 0u, {});
+        return arrow_proxy(std::move(ar), std::move(sc));
```
Klaim, line 30: Static analysis points to this call as not really a move (indeed it copies) but we have different constructors overloads that will not be the same if remove these calls. The warning `performance-move-const-arg` is maybe useful for some case of performance checks but I fear it's pointing at the wrong problem here, it cant check where the moved-form object is passed and if that would change the called overload if we removed these explicit moves. I suspect it will trigger a lot when using `arrow_proxy`. Should we deactivate that check?

Klaim (approved):

test_v01/test_null_array.cpp
JohanMabille: Actually it should move and not copy the structures. I'll fix the implementation of arrow_proxy and we can keep this warning for now.

PR #206, opened 2024-09-10
Add modifiers to arrow_array_schema_proxy

Alex-PLACET:


include/sparrow/array/data_type.hpp
```diff
@@ -443,37 +443,36 @@ namespace sparrow
     }
 
     /// @returns The number of bytes required to store the provided primitive data type.
-    constexpr std::size_t primitive_bytes_count(data_type data_type, int64_t length)
+    constexpr std::size_t primitive_bytes_count(data_type data_type, size_t length)
```
Klaim, line 446: Why this change?
My understanding is that arrow lengths and offsets are all stored in a 64bit integer, so this will break in 32bits?

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -45,32 +64,54 @@ namespace sparrow
         explicit arrow_proxy(ArrowArray&& array, ArrowSchema&& schema);
         /// Constructs an arrow_proxy which takes the ownership of the ArrowArray and uses the provided
         /// ArrowSchema. The array is released when the arrow_proxy is destroyed. The schema is not released.
-        explicit arrow_proxy(ArrowArray&& array, ArrowSchema* schema);
+        explicit arrow_proxy(ArrowArray&& array, ArrowSchema* schema, ownership schema_ownership);
```
Klaim, line 67: I dont like that the ownership has to be specified only for one of the parameters, I would suggest to either request it always, with a default value to shorten code, or to only provide one constructor with pointers and the ownership options.

Also note that [`arrow_data_ownership`](https://github.com/man-group/sparrow/blob/main/include/sparrow/c_interface.hpp#L93) lets the user specify both ownerships already, you should never need to use `ownership` alone in these interfaces.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -45,32 +64,54 @@ namespace sparrow
...
         /// Constructs an arrow_proxy which uses the provided ArrowArray and ArrowSchema.
         /// Neither the array nor the schema are released when the arrow_proxy is destroyed.
-        explicit arrow_proxy(ArrowArray* array, ArrowSchema* schema);
-
-        ~arrow_proxy();
+        explicit arrow_proxy(
```
Klaim, line 70: It should be one [`arrow_data_ownership`](https://github.com/man-group/sparrow/blob/main/include/sparrow/c_interface.hpp#L93) which lets the user specify both ownerships.
Similar to [`external_array_data`](https://github.com/man-group/sparrow/blob/main/include/sparrow/array/external_array_data.hpp#L120), also see [how it looks when the user specifies it](https://github.com/man-group/sparrow/blob/11ef0540b3227b79cf66fb0f331f493ff488b6de/test/external_array_data_creation.hpp#L176-L195).

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,99 +139,176 @@ namespace sparrow
...
-    inline
-    void arrow_proxy::initialize_buffers()
+    inline void arrow_proxy::initialize_buffers()
     {
         m_buffers.clear();
-        const auto buffer_count = static_cast<size_t>(array().n_buffers);
+        const auto buffer_count = static_cast<size_t>(m_array_ref.n_buffers);
         m_buffers.reserve(buffer_count);
-        const data_type data_type = format_to_data_type(schema().format);
+        const enum data_type data_type = format_to_data_type(m_schema_ref.format);
```
Klaim, line 155: Why is `enum` necessary here?

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,99 +139,176 @@ namespace sparrow
...
     }
 
-    inline
-    arrow_proxy::arrow_proxy(ArrowArray&& array, ArrowSchema&& schema)
+    inline arrow_proxy::arrow_proxy(ArrowArray&& array, ArrowSchema&& schema)
         : m_array(std::move(array))
+        , m_array_ref(std::get<1>(m_array))
         , m_schema(std::move(schema))
+        , m_schema_ref(std::get<1>(m_schema))
+        , m_data_ownership({.schema = ownership::owning, .array = ownership::owning})
```
Klaim, line 196:
```suggestion
        , m_data_ownership(owns_arrow_data)
```
See these constants designed specifically for that: 
- [`owns_arrow_data`](https://github.com/man-group/sparrow/blob/11ef0540b3227b79cf66fb0f331f493ff488b6de/include/sparrow/c_interface.hpp#L109C27-L109C43)
- [`doesnt_own_arrow_data`](https://github.com/man-group/sparrow/blob/11ef0540b3227b79cf66fb0f331f493ff488b6de/include/sparrow/c_interface.hpp#L103C27-L103C48)

Also note that it would be best if that is a default parameter (as mentionned in another comment). That makes the ownership clear and irrelevant of how the objects are passed. I dont think it's a good idea in general to have 2 ways to specifies the same ownership, to be clear.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,99 +139,176 @@ namespace sparrow
...
     }
 
-    inline
-    arrow_proxy::arrow_proxy(ArrowArray&& array, ArrowSchema* schema)
+    inline arrow_proxy::arrow_proxy(ArrowArray&& array, ArrowSchema* schema, ownership schema_ownership)
         : m_array(std::move(array))
+        , m_array_ref(std::get<1>(m_array))
         , m_schema(schema)
+        , m_schema_ref(*std::get<0>(m_schema))
+        , m_data_ownership({.schema = schema_ownership, .array = ownership::owning})
```
Klaim, line 209: Same issue as with the other constructor.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,99 +139,176 @@ namespace sparrow
...
+
+        return *this;
+    }
+
+    arrow_proxy::arrow_proxy(arrow_proxy&& other)
+        : m_array(std::move(other.m_array))
+        , m_array_ref(m_array.index() == 0 ? *std::get<0>(m_array) : std::get<1>(m_array))
+        , m_schema(std::move(other.m_schema))
+        , m_schema_ref(m_schema.index() == 0 ? *std::get<0>(m_schema) : std::get<1>(m_schema))
+        , m_data_ownership(other.m_data_ownership)
```
Klaim, line 270: note that you can move it too, it'll copy and if later the ownership data contain something else that's moveable you wont have to maintain that line

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,99 +139,176 @@ namespace sparrow
...
+        if (this == &rhs)
+        {
+            return *this;
+        }
+
+        m_array = std::move(rhs.m_array);
+        m_array_ref = m_array.index() == 0 ? *std::get<0>(m_array) : std::get<1>(m_array);
+        m_schema = std::move(rhs.m_schema);
+        m_schema_ref = m_schema.index() == 0 ? *std::get<0>(m_schema) : std::get<1>(m_schema);
+        m_data_ownership = rhs.m_data_ownership;
```
Klaim, line 291: Same comment, you could move it too.

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: In that case, the proxy always takes the ownership and also hold the data in the variant. The user should not give a misleading information.

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: Ok done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: Ok done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: Ok done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: Ok done

include/sparrow/array/data_type.hpp
Alex-PLACET: Fixed

include/sparrow/arrow_array_schema_proxy.hpp
JohanMabille: I don't understand why we need an additional member for specifying the ownership in the API:
- either the user "moves" the data / schema into the proxy, and therefore transfers the ownership
- or it passes a pointer (because the same structure may be used in different arrow_proxy objects) and therefore keeps the ownership. The structure must be released by the user calling `release` on the "initial" C structure.

I don't see a usecase where the user would like to transfer the owership *and* keep a pointer on the original C structure.

Alex-PLACET: I finally removed the ownership parameter.
According to the Apache Arrow documentation:
https://arrow.apache.org/docs/dev/format/CDataInterface.html#member-allocation

> It is intended for the base structure to be stack- or heap-allocated by the consumer. In this case, the producer API should take a pointer to the consumer-allocated structure.

So, as a consumer, we never get a pointer from a producer third party.

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: removed finally: https://github.com/man-group/sparrow/pull/206#discussion_r1758423297

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: removed finally: https://github.com/man-group/sparrow/pull/206#discussion_r1758423297

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: removed finally: https://github.com/man-group/sparrow/pull/206#discussion_r1758423297

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: removed finally: https://github.com/man-group/sparrow/pull/206#discussion_r1758423297

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -96,99 +135,170 @@ namespace sparrow
...
+    inline arrow_proxy::arrow_proxy(ArrowArray* array, ArrowSchema* schema)
+        : arrow_proxy(array, schema, true)
+    {
+    }
+
+    inline arrow_proxy::arrow_proxy(const arrow_proxy& other)
+        : m_array(&other.m_array_ref)
+        , m_array_ref(*std::get<0>(m_array))
+        , m_schema(&other.m_schema_ref)
+        , m_schema_ref(*std::get<0>(m_schema))
```
JohanMabille, line 235: I think the copy constructor should perform a full copy, including the data.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -96,99 +135,170 @@ namespace sparrow
...
+        m_array_ref = *std::get<0>(m_array);
+        m_schema = &other.m_schema_ref;
+        m_schema_ref = *std::get<0>(m_schema);
+
         validate_array_and_schema();
-        initialize_children();
         initialize_buffers();
+
+        return *this;
+    }
```
JohanMabille, line 257: Same remark here, the assignment operator should copy the data too.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -50,28 +66,44 @@ namespace sparrow
...
+        [[nodiscard]] size_t n_buffers() const;
+        void set_n_buffers(size_t n_buffers);
+        [[nodiscard]] size_t n_children() const;
+        void set_n_children(size_t n_children);
         [[nodiscard]] const std::vector<sparrow::buffer_view<uint8_t>>& buffers() const;
-        [[nodiscard]] const std::vector<arrow_proxy>& children() const;
+        [[nodiscard]] std::vector<sparrow::buffer_view<uint8_t>>& buffers();
+        void set_buffer(size_t index, const buffer_view<uint8_t>& buffer);
+        void set_buffer(size_t index, buffer<uint8_t>&& buffer);
+        [[nodiscard]] std::vector<arrow_proxy> children() const;
```
JohanMabille, line 103: Why returning by values instead of const reference? I don't think we want to allocate a vector each time we need to access the children.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -80,12 +112,17 @@ namespace sparrow
     private:
 
         std::variant<ArrowArray*, ArrowArray> m_array;
+        ArrowArray& m_array_ref;
         std::variant<ArrowSchema*, ArrowSchema> m_schema;
+        ArrowSchema& m_schema_ref;
```
JohanMabille, line 117: `ArrowArray` and `ArrowSchema` should be pointers, otherwise the assignment operators will copy objects instead of reassigning the reference.

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: Done

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +132,180 @@ namespace sparrow
...
         {
-            ArrowArray& array = std::get<1>(m_array);
-            if (array.release != nullptr)
+            return *this;
+        }
+
+        m_array = std::move(rhs.m_array);
+        m_array_ref = m_array.index() == 0 ? *std::get<0>(m_array) : std::get<1>(m_array);
+        m_schema = std::move(rhs.m_schema);
+        m_schema_ref = m_schema.index() == 0 ? *std::get<0>(m_schema) : std::get<1>(m_schema);
```
JohanMabille, line 289: This implementation may leak as `m_array` and `m_schema` are never released while they should. A elegant solution is to swap `m_array` and `rhs.m_array` (same for `schema` and `rhs.m_schema`). Notice that `m_array_ref` and `m_schema_ref` should be pointers and handled the same way. Here the object referenced by `m_schema_ref` is assigned a copy of the content of `m_schema`, which is probably not what you want.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +132,180 @@ namespace sparrow
...
+            return *this;
+        }
+
+        m_array = ArrowArray{};
+        m_array_ref = std::get<1>(m_array);
+        deep_copy_array(other.m_array_ref, other.m_schema_ref, m_array_ref);
+
+        m_schema = ArrowSchema{};
+        m_schema_ref = std::get<1>(m_schema);
+        deep_copy_schema(other.m_schema_ref, m_schema_ref);
```
JohanMabille, line 258: You have the same issue as in the move assign operator, the previous `m_array` and `m_schema` might never be released. A solution is to use the copy and swap idiom:

```cpp
        if (this == &other)
        {
            return *this;
        }

        arrow_proxy tmp(*other);
        swap(*this, tmp); // This assumes the existence of a swap function, you can also swap member by member
        return *this; 
```

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +132,180 @@ namespace sparrow
...
-    void arrow_proxy::initialize_children()
-    {
-        m_children.clear();
-        m_children.reserve(static_cast<std::size_t>(array().n_children));
-        for (int64_t i = 0; i < array().n_children; ++i)
-        {
-            m_children.emplace_back(array().children[i], schema().children[i]);
-        }
-    }
+        std::pair<ArrowSchema, ArrowSchema> copy_data() const;
```
JohanMabille, line 138: This declaration can be removed, it is not called nor implemented.

JohanMabille (changes requested):

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +132,180 @@ namespace sparrow
...
+
+        void update_null_count();
+
+        ArrowArray* copy_array(const ArrowArray& array);
+        ArrowSchema* copy_schema(const ArrowSchema& schema);
```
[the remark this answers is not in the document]
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -223,172 +358,428 @@ namespace sparrow
...
+    inline void arrow_proxy::pop_children(size_t n)
+    {
+        if (!is_created_with_sparrow())
+        {
+            throw arrow_proxy_exception("Cannot set n_buffers on non-sparrow created ArrowArray");
+        }
+
+        if (n > n_children())
+        {
+            throw arrow_proxy_exception("Cannot pop more children than the current number of children");
```
Klaim, line 559: This function is not documented as such. Also, an alternative behavior would be to pop as much as possible (`std::min(n, n_children())`) which is documented would be reasonable too. Anyway, it would be useful to document the prerequisites and behavior for this function.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -80,12 +125,25 @@ namespace sparrow
...
         std::variant<ArrowArray*, ArrowArray> m_array;
+        ArrowArray* m_array_ptr;
         std::variant<ArrowSchema*, ArrowSchema> m_schema;
+        ArrowSchema* m_schema_ptr;
         std::vector<sparrow::buffer_view<uint8_t>> m_buffers;
         std::vector<arrow_proxy> m_children;
+        std::unique_ptr<arrow_proxy> m_dictionary;
+
+        template <typename AA, typename AS>
+        arrow_proxy(AA&& array, AS&& schema, bool);
```
Klaim, line 136: What is that boolean? is it to differentiate that constructor from the public ones? If so consider using a tag type instead, `bool` might be misleading or confusing (in particular at call site).

include/sparrow/arrow_interface/arrow_array.hpp
```diff
@@ -34,23 +39,21 @@ namespace sparrow
      *                  for the the binary or utf-8 view type, which has one additional buffer compared to
      *                  the Columnar format specification (see Binary view arrays). Must be 0 or positive.
      * @param buffers Vector of sparrow::buffer
-     * @param children Vector of arrow_array_shared_ptr representing the children of the ArrowArray.
-     * @param dictionary arrow_array_shared_ptr or nullptr.
+     * @param children Raw vector of ArrowArray pointers or nullptr.
```
Klaim, line 42: `raw vector` is contradictory here, maybe `raw array`, `sequence`, `list` or `range`  would be best.
Same remark for `make_arrow_array_unique_ptr`.

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
```diff
@@ -21,11 +21,50 @@
...
 {
+    /**
+     * Release the children and dictionnary of an ArrowArray or ArrowSchema.
+     * 
+     * @tparam T ArrowArray or ArrowSchema
+     * @param t The ArrowArray or ArrowSchema to release.
+     */
+    template <class T>
+    requires std::same_as<T, ArrowArray> || std::same_as<T, ArrowSchema>
+    void release_common_arrow(T* t)
```
Klaim, line 38: `t` must never be null so that should be a reference. If you prefer a pointer (I prefer a reference), document that it should not be null and add an assertion OR allow it to be null and just check and return if so (and document it).

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
```diff
@@ -21,11 +21,50 @@
...
+#include "sparrow/c_interface.hpp"
 #include "sparrow/utils/memory.hpp"
 #include "sparrow/utils/mp_utils.hpp"
 
 namespace sparrow
 {
+    /**
+     * Release the children and dictionnary of an ArrowArray or ArrowSchema.
+     * 
+     * @tparam T ArrowArray or ArrowSchema
```
Klaim, line 33:
```suggestion
     * @tparam T `ArrowArray` or `ArrowSchema`
```

Keep in mind that doxygen handles markdown syntax.

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
```diff
@@ -21,11 +21,50 @@
...
+     * @param t The ArrowArray or ArrowSchema to release.
```
Klaim, line 34:
```suggestion
     * @param t The `ArrowArray` or `ArrowSchema` to release.
```

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
```diff
@@ -21,11 +21,50 @@
...
+     */
+    template <class T>
+    requires std::same_as<T, ArrowArray> || std::same_as<T, ArrowSchema>
+    void release_common_arrow(T* t)
+    {
+        if (t->dictionary)
+        {
+            if (t->dictionary->release)
+            {
+                t->dictionary->release(t->dictionary);
```
Klaim, line 44: This can throw. As suggested some time ago in another PR, we probably need a "safe invoke" like in mamba to use here and in `arrow_proxy`'s destructor.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -80,12 +125,25 @@ namespace sparrow
     private:
 
         std::variant<ArrowArray*, ArrowArray> m_array;
+        ArrowArray* m_array_ptr;
```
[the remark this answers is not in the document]
Alex-PLACET: I agree that it's maybe a premature optimization. I'm ready to change the code by using a method which implement a visitor or check the variant index, if prefered.

include/sparrow_v01/layout/variable_size_binary_array.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+#include "sparrow_v01/layout/array_base.hpp"
+
+namespace sparrow
+{
+    template <std::ranges::sized_range T, class CR, layout_offset OT = std::int64_t>
+    class variable_size_binary_array;
+
+    /**
+     * @class vs_binary_value_iterator
```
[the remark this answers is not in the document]
Alex-PLACET: Oups sorry, commiting this file was a mistake

include/sparrow_v01/layout/variable_size_binary_array.hpp
```diff
@@ -0,0 +1,319 @@
...
+     *
+     * @brief Iterator over the data values of a variable size binary
+     * layout.
+     *
+     * @tparam L the layout type
+     * @tparam is_const a boolean flag specifying whether this iterator is const.
+     */
+    template <class L, bool is_const>
```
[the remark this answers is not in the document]
Alex-PLACET: Oups sorry, commiting this file was a mistake

include/sparrow_v01/layout/variable_size_binary_array.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+    template <std::ranges::sized_range T, class CR, layout_offset OT>
+    auto variable_size_binary_array<T, CR, OT>::data() const -> const_pointer
+    {
+        return storage().buffers()[1u].template data<const inner_value_type>();
+    }
+
+    template <std::ranges::sized_range T, class CR, layout_offset OT>
+    auto variable_size_binary_array<T, CR, OT>::offset(size_type i) -> offset_iterator
+    {
```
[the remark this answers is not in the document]
Alex-PLACET: Oups sorry, commiting this file was a mistake

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
Alex-PLACET: The exception is not catched here but in the destructor of the `arrow_proxy`

include/sparrow/arrow_interface/arrow_schema.hpp
```diff
@@ -95,37 +108,93 @@ namespace sparrow
...
+        schema->children = children;
+        schema->dictionary = dictionary;
         schema->release = release_arrow_schema;
         return schema;
     };
+
+    /**
+     * Fill the target ArrowSchema with a deep copy of the data from the source ArrowSchema.
+     */
+    inline void deep_copy_schema(const ArrowSchema& source, ArrowSchema& target)
```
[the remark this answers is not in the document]
Alex-PLACET: done

include/sparrow/arrow_interface/arrow_schema.hpp
```diff
@@ -95,37 +108,93 @@ namespace sparrow
...
+        schema->name = private_data->name_ptr();
+        schema->metadata = private_data->metadata_ptr();
+        schema->children = children;
+        schema->dictionary = dictionary;
         schema->release = release_arrow_schema;
         return schema;
     };
+
+    /**
+     * Fill the target ArrowSchema with a deep copy of the data from the source ArrowSchema.
```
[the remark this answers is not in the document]
Alex-PLACET: done

include/sparrow/arrow_interface/arrow_schema.hpp
```diff
@@ -95,37 +108,93 @@ namespace sparrow
...
-                 && std::constructible_from<arrow_schema_private_data::DictionaryType, D>
-    [[nodiscard]] arrow_schema_unique_ptr
-    make_arrow_schema_unique_ptr(F format, N name, M metadata, std::optional<ArrowFlag> flags, C children, D dictionary)
+    [[nodiscard]] arrow_schema_unique_ptr make_arrow_schema_unique_ptr(
+        F format,
+        N name,
+        M metadata,
+        std::optional<ArrowFlag> flags,
+        int64_t n_children,
+        ArrowSchema** children,
```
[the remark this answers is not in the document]
Alex-PLACET: done

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
Alex-PLACET: done

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
Alex-PLACET: Documentation added

include/sparrow/arrow_interface/arrow_array.hpp
```diff
@@ -172,13 +192,90 @@ namespace sparrow
...
+            auto buffer = array.buffers[i];
+            const std::size_t buffer_size = compute_buffer_size(buffer_type, static_cast<size_t>(array.length), static_cast<size_t>(array.offset), data_type);
+            auto* ptr = static_cast<uint8_t*>(const_cast<void*>(buffer));
+            buffers.emplace_back(ptr, buffer_size);
+        }
+        return buffers;
+    }
+
+    /**
+     * Fill the target ArrowArray with a deep copy of the data from the source ArrowArray.
```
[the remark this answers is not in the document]
Alex-PLACET: Done

include/sparrow/arrow_interface/arrow_array.hpp
```diff
@@ -92,64 +97,79 @@ namespace sparrow
...
         int64_t null_count,
         int64_t offset,
         int64_t n_buffers,
         B buffers,
-        int64_t n_children,
-        C children,
-        D dictionary
+        size_t n_children,
+        ArrowArray** children,
+        ArrowArray* dictionary
```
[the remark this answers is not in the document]
Alex-PLACET: Done
children should be not nullptr id n_children is > 0 and dictionary can be nullptr

include/sparrow/arrow_interface/arrow_array.hpp
```diff
@@ -92,64 +97,79 @@ namespace sparrow
...
+        int64_t offset,
+        B buffers,
+        size_t n_children,
+        ArrowArray** children,
+        ArrowArray* dictionary
+    )
+    {
+        SPARROW_ASSERT_TRUE(length >= 0);
+        SPARROW_ASSERT_TRUE(null_count >= -1);
+        SPARROW_ASSERT_TRUE(offset >= 0);
```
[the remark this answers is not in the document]
Alex-PLACET: Done
children should be not nullptr id n_children is > 0 and dictionary can be nullptr

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +153,178 @@ namespace sparrow
...
         {
-            // TODO: Replace with a more specific exception
-            throw std::runtime_error("Invalid ArrowArray format");
+            throw arrow_proxy_exception("Invalid ArrowArray format");
         }
     }
 
-    inline
-    arrow_proxy::arrow_proxy(ArrowArray&& array, ArrowSchema&& schema)
+    template <typename AA, typename AS>
```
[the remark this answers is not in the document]
Alex-PLACET: Done

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +153,178 @@ namespace sparrow
...
 
-    inline
-    void arrow_proxy::validate_array_and_schema() const
+    inline void arrow_proxy::update_children()
     {
-        SPARROW_ASSERT_TRUE(this->array().release != nullptr);
-        SPARROW_ASSERT_TRUE(this->schema().release != nullptr);
+        m_children.clear();
+        m_children.reserve(static_cast<std::size_t>(m_array_ptr->n_children));
+        for (int64_t i = 0; i < m_array_ptr->n_children; ++i)
```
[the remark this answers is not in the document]
Klaim: (discussed, hard to make shorter without `std::range:zip`)

include/sparrow/arrow_array_schema_proxy.hpp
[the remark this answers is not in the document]
Klaim: @JohanMabille Any preference? I tend to prefer what implies less code to debug, but I'm not sure of the cost here.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +153,178 @@ namespace sparrow
...
-    inline
-    void arrow_proxy::initialize_buffers()
+    inline void arrow_proxy::update_buffers()
     {
         m_buffers.clear();
-        const auto buffer_count = static_cast<size_t>(array().n_buffers);
+        const auto buffer_count = static_cast<size_t>(m_array_ptr->n_buffers);
         m_buffers.reserve(buffer_count);
-        const data_type data_type = format_to_data_type(schema().format);
+        const enum data_type data_type = format_to_data_type(m_schema_ptr->format);
```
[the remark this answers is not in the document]
Alex-PLACET: done

include/sparrow/arrow_array_schema_proxy.hpp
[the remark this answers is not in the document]
JohanMabille: If it does not bring significative performance improvement, I agree that less code would be better.

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -95,100 +256,179 @@ namespace sparrow
...
-        for (int64_t i = 0; i < array().n_children; ++i)
-        {
-            m_children.emplace_back(array().children[i], schema().children[i]);
-        }
-    }
+        [[nodiscard]] bool is_arrow_array_valid() const;
+        [[nodiscard]] bool is_arrow_schema_valid() const;
+        [[nodiscard]] bool is_proxy_valid() const;
+
+        static void swap(arrow_proxy& lhs, arrow_proxy& rhs);
```
JohanMabille, line 266: swap is usually implemented a non template method and a free function calling the method on one of its argument. This allows unqualified calls to swap relying on ADL in generic code:

```cpp
template <class T>
void do_something(T& lhs, T& rhs)
{
    // do different things
    using std::swap; // fallback if there is no overload of swap for T
   swap(lhs, rhs);
}
```

JohanMabille (approved):

test/arrow_array_schema_creation.hpp
```diff
@@ -0,0 +1,55 @@
...
+    ArrowSchema schema = *sparrow::make_arrow_schema_unique_ptr(
+                              sparrow::data_type_to_format(sparrow::data_type::UINT8),
+                              "test"sv,
+                              "test metadata"sv,
+                              std::nullopt,
+                              0,
+                              nullptr,
+                              nullptr
+    )
+                              .release();
```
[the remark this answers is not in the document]
Alex-PLACET: Fixed. I added new functions.

include/sparrow/arrow_array_schema_proxy.hpp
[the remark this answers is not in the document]
Alex-PLACET: Fixed

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -14,21 +14,44 @@
...
+    {
+    public:
+
+        explicit arrow_proxy_exception(const std::string& message)
+            : std::runtime_error(message)
+        {
+        }
+    };
+
+    struct arrow_array_and_schema_pointers
```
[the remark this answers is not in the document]
Alex-PLACET: Let's keep it

include/sparrow/arrow_interface/arrow_array_schema_utils.hpp
Alex-PLACET: Done

PR #209, opened 2024-09-17
Removed typed_array, array and external_array

JohanMabille:
This removes all the structures that prevent from changing bottom layers in sparrow; I kept the testing code in case we can reuse part of it later.


DerThorsten:
I would even remove the testing code -- if we want to reuse parts later, we can just use git/github to restore these files

Alex-PLACET (approved):

PR #214, opened 2024-09-19
All binaries all built in the same folder

Alex-PLACET:
Fix the discover and run of the tests executable on Windows (localy).
The issue was that the executable were not able to find the sparrow.dll


Klaim:
This might lead to very nasty surprises when mixing debug and release builds on windows for example (less probable on other platforms but possible too). I would prefer if the binary directory's name was relative to the build mode/configuration name.

JohanMabille (approved):

Alex-PLACET:
@Klaim Ok I added a build type.

Klaim (approved):

PR #224, opened 2024-10-04
Add dictionary encoded array

Alex-PLACET:
For this PR, I had to create a bitmap_offset class. It takes the bitmap and an offset.
As the dictionary uses an index array and a value array, the bitmap already have an offset, and so we can't let the array_base automaticaly add an offset to the iterators.
So now it's the responsability of the array classes to declare the bitmap with and offset and the array_crtp just use the bitmap.begin() method.


include/sparrow/buffer/buffer.hpp
```diff
@@ -151,6 +151,7 @@ namespace sparrow
         constexpr buffer(It first, It last, const A& a = A());
 
         template <std::ranges::input_range Range, allocator A = allocator_type>
+            requires std::same_as<std::ranges::range_value_t<Range>, T>
```
Alex-PLACET, line 154: Without this, we can pass a std::vector<X> to a buffer<Y>. In that case each elements are casted, with possibly some loss.

JohanMabille:
I may have misunderstood the spec, but I don't understand why we need a bitmap and a bitmap_with_offset in the dictionary encoded array. The spec says "The null count of such arrays is dictated only by the validity bitmap of its indices, irrespective of any null values in the dictionary.", so I think the dictionary array should use the bitmap from the index (or keys) arrays. For instance if we have the following data:
- index = [0, 1, 0, 1, null, 2]
- dictionary = ['foo', 'bar', 'baz']
- offset = 0

Then the array "seen" by the user is ['foo', 'bar', 'foo', 'bar', null, 'baz']

Now if offset = 1, the user should see ['bar', 'foo'n 'bar', null, 'baz']

If my assumption is correct, then we don't need the bimap_offset class (the primitive_array owning the indices will already apply its own offset to the bitmap, and if the dicitonary has an offset too, it can add it to the returned iterator in itd own bitmpa_begin method), and we don't need any bitmap member in the dictionary class.

Alex-PLACET:
@JohanMabille The quote you provide is only about the null_count c interface member.
For me, the dictionary layout is the combination of a primitive_array and another array.
In the dictionary layout we don't want to apply any offset to the iterators/element index, as is already handled by the combinations of arrays.
That why I have to modify the CRTP classe to not apply any offset. And so I created a bitmap_offset class.
I agree there is no need of the bitmap_offset for the dictionary class (the offset was already set to 0). I just push a commit to remove it, but we still need a dedicated bitmap for the dictionary as we have to handle nulls of both indexes (keys) and values.
Also I will remove the bitmap_offset class by a subrange, for the primitive and variable binary size.

include/sparrow_v01/layout/array_base.hpp
```diff
@@ -66,7 +64,7 @@ namespace sparrow
      * traits class that must be specialized by array
      * classes inheriting from array_crtp_base.
      */
-    template <class D>
+    template <typename D>
```
DerThorsten, line 67: this makes reviewing harder than it should be.
If you want to make these cosmetic changes make them in a separate PR

Alex-PLACET:
I will create another because the there is too much conflicts with the current merge.

Alex-PLACET:
Replaced by: https://github.com/man-group/sparrow/pull/228

PR #225, opened 2024-10-06
The big merge

JohanMabille:
This merges the refactoring array into the "main" include folder. I can help refactoring the on-going PRs if need. I think this one should be merge asap as it is mostly a mechanical refactoring and it would conflict with any upcoming PR.

Notable changes:
- merges sparrow_v01 into sparrow
- removed old layout classes as they are now replaced with the different array classes
- removed array_data and external_array_data as they were not needed anymore
- clean up the tests
- moved array/data_types and array/data_traits to the new types sublfolder.


DerThorsten (approved):

DerThorsten:
@JohanMabille imho we should merge this as is as fast as possible because any other development does not make sense until this is in

Alex-PLACET (approved):

PR #227, opened 2024-10-07
list-view,big-list-view and fixed-size-list

DerThorsten:
generalized the list layout impl st. all list like layouts (list, list-view, fixed-sized-list) are all implemented with the same core implementation.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -79,7 +79,7 @@ jobs:
 
     - name: Run tests
       working-directory: build
-      run:    ./bin/${{matrix.config.name}}/test_sparrow_lib  # cmake --build . --config ${{matrix.config.name}} --target run_tests_with_junit_report
+      run: cmake --build . --config ${{matrix.config.name}} --target run_tests_with_junit_report
 
     
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -79,7 +79,9 @@ jobs:
 
     - name: Run tests
       working-directory: build
-      run: cmake --build . --config ${{matrix.config.name}} --target run_tests_with_junit_report
+      run:    ./bin/${{matrix.config.name}}/test_sparrow_lib  # cmake --build . --config ${{matrix.config.name}} --target run_tests_with_junit_report
+
+    
 
     - name: Upload test results
       uses: actions/upload-artifact@v4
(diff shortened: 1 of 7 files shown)


.github/workflows/linux.yml
```diff
@@ -79,7 +79,9 @@ jobs:
 
     - name: Run tests
       working-directory: build
-      run: cmake --build . --config ${{matrix.config.name}} --target run_tests_with_junit_report
+      run:    ./bin/${{matrix.config.name}}/test_sparrow_lib  # cmake --build . --config ${{matrix.config.name}} --target run_tests_with_junit_report
```
Alex-PLACET, line 82: If you don't use the cmake target, the report is not generated

include/sparrow/layout/list_layout/list_array.hpp
```diff
@@ -96,78 +148,208 @@ namespace sparrow
         inner_const_reference value(size_type i) const;
         
         // data members
-        flat_array_offset_type * p_list_offsets;
+        //flat_array_offset_type * p_list_offsets;
```
Alex-PLACET, line 151: To remove

include/sparrow/layout/list_layout/list_array.hpp
```diff
@@ -96,78 +148,208 @@ namespace sparrow
...
+    
+    inline fixed_sized_list_array::fixed_sized_list_array(arrow_proxy proxy)
+    :   base_type(std::move(proxy)),
+        m_list_size(0)
+    {
+        // get the list size from the format string
+        const auto format = std::string(this->storage().format());
+        const auto n_digits = format.size() - 3;
+        const auto list_size_str = format.substr(3, n_digits);
+        m_list_size = std::stoull(list_size_str);
```
Alex-PLACET, line 348: Should be extracted in a function which will be used in l.342.
Also it seems that some checks are missing like format.size() >= 3

DerThorsten: with an assert in debug more or a real runtime check?

Alex-PLACET: I would say an assert. But I'm thinking if it should be checked in the proxy. As a constructed proxy should be a valid arrow array + schema.

Alex-PLACET (approved):
LGTM, if you can, please run the clang_format cmake target (you have to define -DACTIVATE_LINTER=ON first)

JohanMabille (approved):

PR #233, opened 2024-10-11
run-length encoded

DerThorsten:


DerThorsten:
I made the data method in the primitive array public since it allows for nice optimizations to have access to it

test/external_array_data_creation.cpp
```diff
@@ -259,5 +259,46 @@ namespace sparrow::test
         arr.release = &release_arrow_array;
     }
 
+    void fill_schema_and_array_for_run_end_encoded(
```
Alex-PLACET, line 262: Can you put this in arrow_array_schema_factory.hpp and arrow_array_schema_proxy_factory.hpp

DerThorsten: for non of the other arrays we did that `external_array_data_creation` has similar code for a bunch of other layouts...

Alex-PLACET: I know but at a moment we will have to use such function to create data ex nihilo

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,260 @@
...
+                    return actual_arr.data();
+                }
+                else if constexpr(std::is_same_v<inner_value_type, std::uint32_t>)
+                {
+                    return actual_arr.data();
+                }
+                else if constexpr(std::is_same_v<inner_value_type, std::uint64_t>)
+                {
+                    return actual_arr.data();
+                }
```
Alex-PLACET, line 172: Can you merge these 3 cases ?

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,260 @@
...
+                else
+                {
+                    throw std::invalid_argument("array type not supported");
```
Alex-PLACET, line 175: Get you replace it by a static assert ?

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    inline auto run_end_encoded_array::operator[](std::uint64_t i) const -> array_traits::const_reference
+    {
+        // visit the variant
+        auto ret =visit(
+            [i, this](const auto& acc_lengths_ptr) -> array_traits::const_reference
+            {
+                
+                auto it = std::upper_bound(
```
Alex-PLACET, line 210: const too

test/external_array_data_creation.cpp
DerThorsten: I dont know what ex nihilo means =)

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
DerThorsten: not sure what you mean

test/external_array_data_creation.cpp
Alex-PLACET: I mean at some point we will have to create C structures when the user will want to create an array from some buffers. We want nice functions to create c structures yes.

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
Alex-PLACET: I mean the variable can be const.

DerThorsten:
@JohanMabille  on windows I see this `'sparrow::array_wrapper::wrapper_ptr' to 'sparrow::array_wrapper *'` since I added arrays using  child arrays to the factory / dispatch switches.

Not sure if I am doing smth wrong or if there is still an issue with the wrapper class

include/sparrow/layout/primitive_array.hpp
```diff
@@ -18,10 +18,13 @@
 #include "sparrow/layout/array_base.hpp"
 #include "sparrow/utils/iterator.hpp"
 #include "sparrow/utils/nullable.hpp"
-
+//#include "sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp"
```
Alex-PLACET, line 21: To remove ?

src/run_encoded_array.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+    };
+    template<>
+    struct is_usable_array<primitive_array<std::uint32_t>> : std::true_type
+    {
+    };
+    template<>
+    struct is_usable_array<primitive_array<std::uint64_t>> : std::true_type
+    {
+    };
```
Alex-PLACET, line 25:
```suggestion
    template<class T>
    concept usable_array =
             mpl::is_type_instance_of_v<T, primitive_array> && (
             std::same_as<typename T::value_type, std::uint16_t> ||
             std::same_as<typename T::value_type, std::uint32_t> ||
             std::same_as<typename T::value_type, std::uint64_t>);
    };
```
or something similar

include/sparrow/layout/array_wrapper.hpp
```diff
@@ -148,7 +168,8 @@ namespace sparrow
     template <class T>
     constexpr enum data_type array_wrapper_impl<T>::get_data_type() const
     {
-        return arrow_traits<typename T::inner_value_type>::type_id;
+        return detail::get_data_type_from_array<T>::get();
+        //return arrow_traits<typename T::inner_value_type>::type_id;
```
Alex-PLACET, line 172: To remove ?

src/run_encoded_array.cpp
```diff
@@ -0,0 +1,73 @@
...
+            [](const auto& actual_arr) -> acc_length_ptr_variant_type
+            {
+                using array_type = std::decay_t<decltype(actual_arr)>;
+                if constexpr( is_usable_array<array_type>::value)
+                {
+                    return actual_arr.data();
+                }
+                else
+                {
+                    throw std::invalid_argument("array type not supported");
```
Alex-PLACET, line 44: To replace with a static assert

src/run_encoded_array.cpp
DerThorsten: it should be inner_value_type, but your suggestions are fine with me

include/sparrow/layout/array_wrapper.hpp
DerThorsten: sure

src/run_encoded_array.cpp
DerThorsten: this is not working...even with a dependent false this will aways result in a compile-time-error

Alex-PLACET: It's really unclear to my why, sometimes, a simple "else" is enough, sometimes we have to use a "dependent false", and sometimes nothing works.
Have you tried to do an "else if (!is_usable_array ...)"

DerThorsten: so a non-templated (library) function, I believe it will never work because its "fully evaluated"

Alex-PLACET: Ho you right, I didn't noticed that

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    template<bool CONST>
+    typename array_traits::const_reference run_encoded_array_iterator<CONST>::dereference() const
+    {
+        return array_element(*p_encoded_values_array, static_cast<std::size_t>(m_run_end_index));
+    }
+
+    class SPARROW_API run_end_encoded_array final
```
JohanMabille, line 109: No need for `final` here.

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,222 @@
...
+    typename array_traits::const_reference run_encoded_array_iterator<CONST>::dereference() const
+    {
+        return array_element(*p_encoded_values_array, static_cast<std::size_t>(m_run_end_index));
+    }
+
+    class SPARROW_API run_end_encoded_array final 
+    {
+    public:
+        using self_type = run_end_encoded_array;
+        using base_type = array_crtp_base<self_type>;
```
JohanMabille, line 113: This can be removed now.

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,222 @@
...
+    // Ie nullabes values, not values !!! 
+    template<bool CONST>
+    class run_encoded_array_iterator : public iterator_base<
+        run_encoded_array_iterator<CONST>,
+        array_traits::const_reference,
+        std::forward_iterator_tag,
+        array_traits::const_reference
+    >
+    {   
+        private:
```
JohanMabille, line 56: Nitpicking: indentation (same for the other sections below).

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,222 @@
...
+        };
+    }
+
+    // this iteratas over the **actual** values of the run encoded array
+    // Ie nullabes values, not values !!! 
+    template<bool CONST>
+    class run_encoded_array_iterator : public iterator_base<
+        run_encoded_array_iterator<CONST>,
+        array_traits::const_reference,
+        std::forward_iterator_tag,
```
JohanMabille, line 52: Why not mankig it random_access? Is there anything preventing it?

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+        arrow_proxy m_proxy;
+        std::uint64_t m_encoded_length;
+        
+        cloning_ptr<array_wrapper> p_acc_lengths_array;
+        cloning_ptr<array_wrapper> p_encoded_values_array;
+        acc_length_ptr_variant_type m_acc_lengths;
+
+        // friend classes
+        friend class array_crtp_base<self_type>;
```
JohanMabille, line 149: This is not required anymore.

JohanMabille (commented):
Except if you plan to add other files under the `run_end_encoded_layout` subfolder, I think the `run_end_encoded_array` should be moved under the `layout` folder.

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
DerThorsten: random access would be not O(N) but O(logn)

JohanMabille:
Regarding the Windows link DLL, it should be possible to export the methods of `run_encoded_array` instead of the whole class, as described [here](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4251?view=msvc-170&redirectedfrom=MSDN). This would avoid the need for exporting `arrow_proxy` and potentially other data members.

Not sure if it is the strategy we want to adopt on the long run, but it should be neough to make this PR pass the CI.

Alex-PLACET (approved):

JohanMabille (approved):

PR #236, opened 2024-10-14
Refactored dictionary_encoded_array

### LINKED ISSUE #193: Refactoring: port the dictionary encoded layout to the array impl hierarchy

JohanMabille:
Fixes #193 

- the new "implementation" of dictionary_encoded_array's iterator is not optimal: it could store an iterator on the keys array instead of accessing it through an index. This requires refactoring the `functor_index_iterator` to avoid code duplication (this iterator works the same sohuld it store an index or a subiterator), that I prefer to do in a dedicated PR to avoid "noise" in this PR.
- iterators on the bitmap and the values can be added back later:
  - this won't affect the current impementation of the element iterator (this one won't become a layout_iterator based on bitmap_iterator and value_iterator as it will complexify the code and degrade performance by introducing many visit in the access operator)
  - this requires to design a variant-like class that can hold references, unlike `std::variant`


src/list_value.cpp
```diff
@@ -33,4 +34,14 @@ namespace sparrow
...
     }
+
+    bool operator==(const list_value& lhs, const list_value& rhs)
+    {
+        bool res = lhs.size() == rhs.size();
+        for (std::size_t i = 0; res && i < lhs.size(); ++i)
+        {
+            res = lhs[i] == rhs[i];
+        }
+        return res;
```
Alex-PLACET, line 45:
```suggestion
        return std::ranges::equal(lhs, rhs);
```

src/struct_value.cpp
```diff
@@ -32,4 +33,14 @@ namespace sparrow
...
     }
+
+    bool operator==(const struct_value& lhs, const struct_value& rhs)
+    {
+        bool res = lhs.size() == rhs.size();
+        for (std::size_t i = 0; res && i < lhs.size(); ++i)
+        {
+            res = lhs[i] == rhs[i];
+        }
+        return res;
```
Alex-PLACET, line 44: return std::ranges::equal(lhs, rhs);

src/list_value.cpp
JohanMabille: `list_value` is not a range yet.

src/struct_value.cpp
JohanMabille: `struct_value` is not a range yet.

test/test_dispatch.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 
+#include "sparrow/layout/array_helper.hpp"
```
Alex-PLACET, line 16: Is it used ?

src/list_value.cpp
Alex-PLACET: Can you add a todo ?

DerThorsten (changes requested):
the array needs to be added [the array_factory ](https://github.com/man-group/sparrow/blob/main/src/array_factory.cpp)

JohanMabille:
> the array needs to be added [the array_factory ](https://github.com/man-group/sparrow/blob/main/src/array_factory.cpp)

Yep but that requires a change in the return_type of the data_type method of `array_wrapper`. I wanted to avoid conflicts with your incoming PR about run_encoded_layout and do it later. I can wait for your PR to be merge, rebase and add it though.

test/test_dictionary_encoded_array.cpp
```diff
@@ -58,6 +59,11 @@ namespace sparrow
         return ar;
     }
 
+    nullable<std::string_view> get_dict_value(layout_type_cref r)
```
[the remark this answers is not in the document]
JohanMabille: Not here, nullable acts as a reference proxy. When the `value_type` of an array is `nullable<T>`, the `const_reference` is usually `nullable<const T&>` (there are some exceptions like `std::string` or `null_type`). For Arays returning a variant, the const_reference is a variant of nullable (objects, because variant cannot hold references) holding references.

test/test_dispatch.cpp
JohanMabille: Yes, the `array_element` function is declared in `array_helper.hpp` now.

JohanMabille:
The last commit is just the rebase, I still need to plug the dictionary_encoded_array into the dispatch and array_factory.

Alex-PLACET (approved):
LGTM

DerThorsten (approved):

PR #237, opened 2024-10-15
Minimal implementation of the array facade

JohanMabille:


src/array.cpp
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+#include "sparrow/array.hpp"
+#include "sparrow/array_factory.hpp"
+#include "sparrow/layout/array_helper.hpp"
+
+namespace sparrow
+{
+    array::array(ArrowArray&& array, ArrowSchema&& schema)
+        : p_array(array_factory(arrow_proxy(std::move(array), std::move(schema))))
```
Alex-PLACET, line 22:
```suggestion
        : p_array(array_factory(arrow_proxy(std::forward<ArrowArray>(array), std::forward<ArrowSchemay>(schema))))
```

src/array.cpp
```diff
@@ -0,0 +1,46 @@
...
+    {
+    }
+
+    array::array(ArrowArray&& array, ArrowSchema* schema)
+        : p_array(array_factory(arrow_proxy(std::move(array), schema)))
```
Alex-PLACET, line 27:
```suggestion
        : p_array(array_factory(arrow_proxy(std::forward<ArrowArray>(array), schema)))
```

Alex-PLACET (approved):
LGTM

src/array.cpp
DerThorsten: `std::move` is correct here.
`std::forward` is for generic code, but here are not templates

src/array.cpp
DerThorsten: `std::move` is correct here.
`std::forward` is for generic code, but here are not templates

DerThorsten (approved):

PR #239, opened 2024-10-16
Move arrow_array arrow_schema and arrow_array_schema_proxy implem to …

Alex-PLACET:
…cpp files

To fix issues with the release_arrow_schema and release_arrow_array adresses.
The code is copy pasted + SPARROW_API and some includes. There is no modification in the impl


test/test_arrow_array.cpp
```diff
@@ -282,21 +283,21 @@ TEST_SUITE("C Data Interface")
         {
```
Alex-PLACET, line 283: Starting from this point it's only formatting

DerThorsten (approved):

PR #241, opened 2024-10-16
Implemented arrow data getter / extraction

JohanMabille:


test/test_array.cpp
```diff
@@ -72,6 +72,121 @@ namespace sparrow
...
+                ArrowArray ar{};
+                test::fill_schema_and_array<scalar_value_type>(sc, ar, size, offset, {});
+                array a(std::move(ar), std::move(sc));
+
+                ArrowSchema* sc_ptr = nullptr;
+                ArrowArray* ar_ptr = nullptr;
+                a.get_arrow_array(ar_ptr).get_arrow_schema(sc_ptr);
+
+                auto pa = primitive_array<scalar_value_type>(arrow_proxy(ar_ptr, sc_ptr));
+
```
Alex-PLACET, line 143: You have to call the release callback of both array and schema and delete both pointers.

test/test_array.cpp
```diff
@@ -72,6 +72,121 @@ namespace sparrow
...
+                test::fill_schema_and_array<scalar_value_type>(sc, ar, size, offset, {});
+                array a(&ar, &sc);
+
+                ArrowSchema* sc_ptr = nullptr;
+                ArrowArray* ar_ptr = nullptr;
+                a.get_arrow_array(ar_ptr).get_arrow_schema(sc_ptr);
+
+                auto pa = primitive_array<scalar_value_type>(arrow_proxy(ar_ptr, sc_ptr));
+
+                CHECK_EQ(pa, pa_ctrl);
```
Alex-PLACET, line 128: You have to call the release callback of both array and schema and delete both pointers.

Alex-PLACET (changes requested):

test/test_array.cpp
JohanMabille: `ar_ptr` points to `&ar` and `sc_ptr` points to `&sc` so they should not be deleted (`ar` and `sc` are allocated on the stack and will be destroyed at the end of the scope). The release methods of these structures must still be called though, I find it more intuitive to call them on the original structure (i.e. `ar` and `sc`).

test/test_array.cpp
JohanMabille: ar_ptr points to &ar and sc_ptr points to &sc so they should not be deleted (ar and sc are allocated on the stack and will be destroyed at the end of the scope). The release methods of these structures must still be called though, I find it more intuitive to call them on the original structure (i.e. ar and sc).

Alex-PLACET: You are right for this one the ar_ptr and sc_ptr are pointers over ar and sc.

test/test_array.cpp
Alex-PLACET: I mixed with extrat. You are right, a will release the array and schema.

Alex-PLACET (approved):

PR #247, opened 2024-10-22
Move semantic

JohanMabille:
Simplification of arrow_proxy move constructor and move semantic of composed layouts for (almost) free ;)


DerThorsten (approved):
lgtm

PR #249, opened 2024-10-24
Constructorsv2

DerThorsten:


include/sparrow/layout/primitive_array.hpp
```diff
@@ -130,6 +176,106 @@ namespace sparrow
...
+            }
+            else{
+                // set bit to 0
+                bitmap_ptr[i / 8] &= ~(1 << (i % 8));
+            }
+            ++value_iter;
+            ++is_non_null_iter;
+        }
+        arrow_proxy proxy{std::move(arr), std::move(schema)};
+        return proxy;
```
Alex-PLACET, line 237: Please use arrow_array_schema_factory.hpp instead

include/sparrow/array.hpp
```diff
@@ -47,9 +55,22 @@ namespace sparrow
...
+        friend detail::array_access;
     };
+
+    template<class ARRAY_TYPE>
+    requires std::is_rvalue_reference_v<ARRAY_TYPE&&>
+    array::array(ARRAY_TYPE&& array)
+    {   
+        auto storage = detail::array_access::extract_arrow_proxy(std::move(array));
+        p_array = array_factory(std::move(storage));
+    }
```
JohanMabille, line 73: Since ARRAY_TYPE is known, maybe something like:
```cpp
    template<class ARRAY_TYPE>
    requires std::is_rvalue_reference_v<ARRAY_TYPE&&>
    array::array(ARRAY_TYPE&& array)
        : p_array(new array_wrapper_impl<ARRAY_TYPE>(std::move(array)))
    {
    }
```

Would be enough? It avoids extracting the proxy and going through the factory.

include/sparrow/array.hpp
```diff
@@ -47,9 +55,21 @@ namespace sparrow
         SPARROW_API size_type size() const;
         SPARROW_API const_reference operator[](size_type) const;
 
+        SPARROW_API cloning_ptr<array_wrapper> && extract_array_wrapper() &&;
```
JohanMabille, line 58: The return type should be a value type instead of an rvalue reference to prevent dangling references.

include/sparrow/layout/array_access.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+        template<class ARRAY>
+        static inline sparrow::arrow_proxy& storage(ARRAY& array)
+        {
+            return array.storage();
+        }
+
+        template<class ARRAY>
+        requires(std::is_rvalue_reference_v<ARRAY&&>)
+        static inline sparrow::arrow_proxy&& extract_arrow_proxy(ARRAY&& array)
```
JohanMabille, line 40: Same remark here (notice this is just about the return type, the implementation is correct).

include/sparrow/layout/array_base.hpp
```diff
@@ -119,13 +122,17 @@ namespace sparrow
...
 
         const_bitmap_iterator bitmap_begin() const;
         const_bitmap_iterator bitmap_end() const;
 
-        const_bitmap_iterator bitmap_cbegin() const;
-        const_bitmap_iterator bitmap_cend() const;
+
+    private:
+
+        arrow_proxy && extract_arrow_proxy() &&;
```
JohanMabille: Same remark here.

include/sparrow/layout/list_layout/list_array.hpp
```diff
@@ -491,4 +502,44 @@ namespace sparrow
         const auto offset = i * m_list_size;
         return std::make_pair(offset, offset + m_list_size);
     }
+
+
+    inline arrow_proxy fixed_sized_list_array::create_proxy(std::uint64_t list_size, array && flat_values)
```
JohanMabille, line 507: I think we should allow to pass a range for the null values (similar to what we do [here](https://github.com/man-group/sparrow/blob/main/test/external_array_data_creation.cpp#L113-L121))

include/sparrow/layout/primitive_array.hpp
```diff
@@ -155,6 +192,120 @@ namespace sparrow
...
+                bitmap_ptr[i / 8] |= static_cast<std::uint8_t>(1 << (i % 8));
+            }
+            else
+            {
+                // set bit to 0
+                bitmap_ptr[i / 8] &= static_cast<std::uint8_t>(~(1 << (i % 8)));
+            }
+            ++value_iter;
+            ++is_non_null_iter;
+        }
```
JohanMabille, line 238: Filling the bitmap should be done in a dedicated method (notice it is alrady implemented [here](https://github.com/man-group/sparrow/blob/main/include/sparrow/arrow_interface/arrow_array_schema_factory.hpp#L57-L71)) to avoid code duplictation.

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -78,6 +79,7 @@ namespace sparrow
         SPARROW_API static acc_length_ptr_variant_type get_acc_lengths_ptr(const array_wrapper& ar);
         SPARROW_API std::uint64_t get_run_length(std::uint64_t run_index) const;
 
+        arrow_proxy && extract_arrow_proxy() &&;
```
JohanMabille, line 82: The return type should be a value instead of an rvalue reference.

JohanMabille (commented):
I don't see the constructor for variable_size_binary_layout, is that a leftover or is it intentional?

include/sparrow/layout/list_layout/list_array.hpp
DerThorsten: agree!

DerThorsten:
> I don't see the constructors for `variable_size_binary_array` and `struc_array_array`, is that a leftover or is it intentional?

I only implemented the primitive array  and **a single nested array** (only the fixed_size_binary) because it made sense to show this to you before implementing **everything**

include/sparrow/layout/primitive_array.hpp
DerThorsten: @JohanMabille the function you show does something different.
I pass a complete range of non-null / null (ie a range as long as the values filled with true and false).
The API of the function you show has a list of positions where values are missing.

This is also fine with me to prefer such API but they are 2 different things

include/sparrow/layout/primitive_array.hpp
```diff
@@ -155,6 +192,120 @@ namespace sparrow
...
+            if(is_non_null_val)
+            {
+                data_ptr[i] = static_cast<T>(*value_iter);
+                // set bit to 1
+                bitmap_ptr[i / 8] |= static_cast<std::uint8_t>(1 << (i % 8));
+            }
+            else
+            {
+                // set bit to 0
+                bitmap_ptr[i / 8] &= static_cast<std::uint8_t>(~(1 << (i % 8)));
```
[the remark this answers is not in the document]
DerThorsten: this does not work in the generic case.
An iterator might might not be allowed to be dereferenced in the case of a missing value

include/sparrow/layout/primitive_array.hpp
```diff
@@ -155,6 +192,120 @@ namespace sparrow
...
+    template <std::ranges::input_range VALUE_RANGE, std::ranges::input_range BOOL_RANGE>
+    requires std::convertible_to<std::ranges::range_value_t<VALUE_RANGE>, T> &&
+             std::convertible_to<std::ranges::range_value_t<BOOL_RANGE>, bool>
+    arrow_proxy primitive_array<T>::create_proxy(VALUE_RANGE && values, BOOL_RANGE && is_non_null)
+    {
+
+        const auto range_size = std::ranges::size(values);
+
+        // buffers
+        auto bitmap_ptr = new std::uint8_t[(range_size + 7) / 8];
```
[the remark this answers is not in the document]
DerThorsten: tbh I dont see the point of sparrow::buffer at at all...

include/sparrow/layout/primitive_array.hpp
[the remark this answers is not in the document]
DerThorsten: > You should set the value anyway, in case the bitmap is changed later.

if someone changes the value of the bitmap, one has change the value anyhow

> In the case the user want to set to null a single element in 100 000 elements, he has to create a vector of 100 000 elements or a range which return false for this index ?

It does not need to be a vector at all, it can be any iterator. Second, there is an overload where you just pass the values (this then generates an iterator always returning true (I trust compilers do do some sort of optimization st. this will not be very bad)) and then set by hand change the bitmap later on.
Furthermore, the price of an N element bool-vector would be 1/8 of the size of the actual values in the worst case.

include/sparrow/layout/primitive_array.hpp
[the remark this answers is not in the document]
DerThorsten: I could use the buffer...but me personally, I prefer to be explicit here and just allocate the ptrs by hand.
But I am fine with changing it so some buffer

include/sparrow/layout/primitive_array.hpp
JohanMabille: > This is also fine with me to prefer such API but they are 2 different things

Ah indeed. I think passing a range with the indices of the missing values would be more convenient.

include/sparrow/layout/primitive_array.hpp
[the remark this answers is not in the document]
Alex-PLACET: > And if you set a missing value to non-missing you better change the value anyhow
I don't know the use cases, but there is no cons and only benefits to copy "null" values.

include/sparrow/layout/primitive_array.hpp
DerThorsten: > > This is also fine with me to prefer such API but they are 2 different things
> 
> Ah indeed. I think passing a range with the indices of the missing values would be more convenient.

My thinking was the following:
When you have data, do you rather have a list of where the data is missing, or do you have some sort of functor which takes in  a single data point and then decides if its missing or not. For instance once could think of a measurements of floats, and any negative number indicates that something went wrong, then you want to exclude this data point -- in such cases I would prefer to pass an iterator to bools instead of indices)

include/sparrow/layout/primitive_array.hpp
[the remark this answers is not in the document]
DerThorsten: > I don't know the use cases, but there is no cons and only benefits to copy "null" values.

Again, the value iterator might be undefined for the places where data is missing.
Assume you have `std::vector<sparrow::nullable<float>>` and the values are a transform iterator and also the missing-vs-non missing is a transform-iterator.

In that case you will get an exception if you try to access `.value()` at places where you don't have data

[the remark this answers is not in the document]
Alex-PLACET: You have to use the get() function in that case

include/sparrow/layout/primitive_array.hpp
JohanMabille: The convenient thing with indices of missing values is that you can provide a default empty argument, and omit it when you know that you don't have any missing values. And building a range of indices from the full bitmap range could be quite easy with the `std::range` functions, I'm not sure the opposite would be as easy (but I might have missed some range constructions).

include/sparrow/layout/primitive_array.hpp
```diff
@@ -155,6 +192,120 @@ namespace sparrow
...
     }
 
+    template <class T>
+    template <std::ranges::input_range VALUE_RANGE, std::ranges::input_range BOOL_RANGE>
+    requires std::convertible_to<std::ranges::range_value_t<VALUE_RANGE>, T> &&
+             std::convertible_to<std::ranges::range_value_t<BOOL_RANGE>, bool>
+    arrow_proxy primitive_array<T>::create_proxy(VALUE_RANGE && values, BOOL_RANGE && is_non_null)
+    {
+
+        const auto range_size = std::ranges::size(values);
```
Alex-PLACET, line 202: values size must be equal to is_non_null size

include/sparrow/layout/primitive_array.hpp
DerThorsten: > The convenient thing with indices of missing values is that you can provide a default empty argument, and omit it when you know that you don't have any missing values. 

jeah that's indeed a good argument

include/sparrow/buffer/dynamic_bitset/dynamic_bitset.hpp
```diff
@@ -17,6 +17,8 @@
 #include "sparrow/buffer/dynamic_bitset/dynamic_bitset_base.hpp"
 #include "sparrow/buffer/buffer.hpp"
 
+#include <iostream>
```
JohanMabille, line 20: Looks like a leftover

Alex-PLACET (approved):

include/sparrow/buffer/dynamic_bitset/dynamic_bitset.hpp
```diff
@@ -92,4 +109,81 @@ namespace sparrow
...
+    template <class T>
+    concept validity_bitmap_input = 
+        std::same_as<T, validity_bitmap&&> ||
+        std::same_as<T, const validity_bitmap&> ||
+        (std::ranges::input_range<T> && std::same_as<std::ranges::range_value_t<T>, bool>) ||
+        (std::ranges::input_range<T> && std::unsigned_integral<std::ranges::range_value_t<T>> );
+
+
+    template <validity_bitmap_input R>
+    validity_bitmap ensure_validity_bitmap(std::size_t size, R&& validity_input)
```
JohanMabille, line 184: If you pass an `validity_input` as an rvalue reference, R will be deduced as the "raw type R", not as the rvalue reference.
So the `validity_bitmap_input` concept should rather be defined as:
```cpp
    template <class T>
    concept validity_bitmap_input = 
        std::same_as<T, validity_bitmap> || // Not validity_bitmap&&
        std::same_as<T, const validity_bitmap&> ||
        (std::ranges::input_range<T> && std::same_as<std::ranges::range_value_t<T>, bool>) ||
        (std::ranges::input_range<T> && std::unsigned_integral<std::ranges::range_value_t<T>> );
```

include/sparrow/layout/array_access.hpp
JohanMabille: The return type should be `sparrow::arrow_proxy` instead.

include/sparrow/layout/array_base.hpp
```diff
@@ -108,17 +109,19 @@ namespace sparrow
...
-        array_crtp_base(arrow_proxy);
+        explicit array_crtp_base(arrow_proxy);
 
         array_crtp_base(const array_crtp_base&) = default;
         array_crtp_base& operator=(const array_crtp_base&) = default;
 
         array_crtp_base(array_crtp_base&&) = default;
         array_crtp_base& operator=(array_crtp_base&&) = default;
 
+        [[nodiscard]] arrow_proxy && extract_arrow_proxy() &&;
```
JohanMabille, line 120: The return type should be `arrow_proxy`

JohanMabille (changes requested):

include/sparrow/layout/array_base.hpp
DerThorsten: jup, missed that one!

JohanMabille (approved):

PR #257, opened 2024-10-28
Record batch

JohanMabille:


include/sparrow/record_batch.hpp
```diff
@@ -0,0 +1,239 @@
...
+{
+    /**
+     * Table-like data structure.
+     *
+     * A record batch is a collection of equal-length arrays mapped to
+     * names.
+     *
+     * @tparm T the type of the names
+     */
+    template <class T = std::string>
```
DerThorsten, line 37: do we really need string as a template here?

src/array.cpp
```diff
@@ -81,5 +81,23 @@ namespace sparrow
...
+        {
+            return rhs.visit([&typed_lhs](const auto& typed_rhs) -> bool
+            {
+                if constexpr (!std::same_as<decltype(typed_lhs), decltype(typed_rhs)>)
+                {
+                    return false;
+                }
+                else
+                {
+                    return typed_lhs == typed_rhs;
```
DerThorsten, line 97: I am a bit puzzled by this. Is == implemented by each layout?

include/sparrow/record_batch.hpp
JohanMabille: Probably not, I'll remove it.

src/array.cpp
JohanMabille: Layouts inheriting from `array_crtp_base` get it for free. The other layouts must implement it, I'll check if it's the case for each of them.

test/test_record_batch.cpp
Alex-PLACET, line 1: Some test are missings for:
size_type nb_columns() const;
size_type nb_rows() const;
name_range names() const;
column_range columns() const;

JohanMabille: The missing tests have been added.

src/array.cpp
JohanMabille: Checked, all the layouts inherit or implement `operator==`.

Alex-PLACET (approved):

DerThorsten (approved):

PR #259, opened 2024-10-29
convinience constructors for list and list view

DerThorsten:


include/sparrow/layout/layout_utils.hpp
```diff
@@ -73,4 +76,29 @@ namespace sparrow::detail
...
+} // namespace sparrow::detail
+
+
+template<class OFFSET_TYPE>
+concept offset_type = std::is_same<std::remove_const_t<OFFSET_TYPE>, std::uint32_t>::value || 
+                        std::is_same<std::remove_const_t<OFFSET_TYPE>, std::uint64_t>::value;
+
+template<offset_type OFFSET_TYPE, std::ranges::range SIZES_RANGE>
+requires(std::unsigned_integral<std::ranges::range_value_t<SIZES_RANGE>>)
+u8_buffer<OFFSET_TYPE> offset_buffer_from_sizes(SIZES_RANGE && sizes)
```
JohanMabille, line 88: Is that specific to the different list layouts? If so, it should be moved to `list_array.hpp`. In any case I think it can be put in the `detail` namespace.

include/sparrow/layout/list_layout/list_array.hpp
```diff
@@ -190,8 +192,19 @@ namespace sparrow
         list_array_impl(self_type&&) = default;
         list_array_impl& operator=(self_type&&) = default;
 
+        template<class ... ARGS>
+        requires(mpl::excludes_copy_and_move_ctor_v<list_array_impl<BIG>, ARGS...>)
+        list_array_impl(ARGS && ... args): self_type(create_proxy(std::forward<ARGS>(args)...))
+        {}
```
JohanMabille, line 198: Do we want to allow constructions such as:
```cpp
list_array arr = {{1, 3, 4}, {2, 4}, {5, 1 ,9}};
```
?

include/sparrow/layout/layout_utils.hpp
DerThorsten: other layouts need offsets from sizes I believe

include/sparrow/layout/list_layout/list_array.hpp
DerThorsten: I don't think so. They are very special in the sense that they only work if the child array is a primitive array.

DerThorsten: on the other hand it might be a common use case...
I am unsure

JohanMabille: Ok let's see if users ask for it in the future.

JohanMabille (approved):

Alex-PLACET (approved):

PR #264, opened 2024-10-31
Added some documentation

JohanMabille:


.github/workflows/linux.yml
```diff
@@ -20,10 +20,10 @@ jobs:
       fail-fast: false
       matrix:
         sys:
-        - {image: ubuntu-22.04, compiler: clang, version: '16', config-flags: '', stdlib: 'libstdc++-12', date-polyfill: 'ON'}
-        # - {compiler: clang, version: '16', config-flags: '-DCMAKE_CXX_FLAGS=-stdlib=libc++', stdlib: 'libc++-17', date-polyfill: 'ON' }
         - {image: ubuntu-22.04, compiler: clang, version: '17', config-flags: '', stdlib: 'libstdc++-12', date-polyfill: 'ON'}
         # - {compiler: clang, version: '17', config-flags: '-DCMAKE_CXX_FLAGS=-stdlib=libc++', stdlib: 'libc++-17', date-polyfill: 'ON' }
+        - {image: ubuntu-22.04, compiler: clang, version: '18', config-flags: '', stdlib: 'libstdc++-12', date-polyfill: 'ON'}
```
Alex-PLACET, line 25: CAn you use ubuntu 24, it will be faster as the default llvm version is also 18

include/sparrow/array_api.hpp
Alex-PLACET, line 1: Can you run clang format ?

DerThorsten (approved):

src/array.cpp
```diff
@@ -52,7 +52,15 @@ namespace sparrow
...
         return p_array->get_arrow_proxy();
     }
-    
+
+    /**
+     * Compares the content of two arrays.
+     *
+     * @param lhs the first array to compare
+     * @param rhs the second array to compare
+     * @return \c true if the contesnts of the tow arrays
```
Alex-PLACET, line 61:
```suggestion
     * @return \c true if the contents of both arrays
```

src/array.cpp
```diff
@@ -52,7 +52,15 @@ namespace sparrow
     {
         return p_array->get_arrow_proxy();
     }
-    
+
+    /**
```
Alex-PLACET, line 56: Should be in the hpp

include/sparrow/array_api.hpp
```diff
@@ -33,28 +43,96 @@ namespace sparrow
...
+         */
         SPARROW_API size_type size() const;
+
+        /**
+         * Returns a constant reference to the element at specified \c index.
+         *
+         * @param index The position of the element in the array.
+         * @return A constant reference to the element.
+         */
         SPARROW_API const_reference operator[](size_type) const;
```
Alex-PLACET, line 122: can you add `index` in the signature
```suggestion
        SPARROW_API const_reference operator[](size_type index) const;
```

include/sparrow/array_api.hpp
```diff
@@ -33,28 +43,96 @@ namespace sparrow
...
+         *
+         * @param array The ArrowArray structure to transfer into the array.
+         * @param schema The ArrowSchema to reference in the array.
+         */
         SPARROW_API array(ArrowArray&& array, ArrowSchema* schema);
+
+        /**
+         * Constructs an array from the given Arrow C structures. Both structures
+         * are referenced from the array and can still be used normally after calling
+         * this cnstructor.
```
Alex-PLACET, line 104:
```suggestion
         * this constructor.
```

include/sparrow/array_api.hpp
```diff
@@ -33,28 +43,96 @@ namespace sparrow
...
+         * @param a A shared pointer holding the layout.
+         */
         template <layout A>
-        array(std::shared_ptr<A> a);
-
+        explicit array(std::shared_ptr<A> a);
+
+        /**
+         * Constructs an array from the given Arrow C structures, whose ownerhips
+         * is tranferred to the array. The user should not use \c array nor \c schema
```
Alex-PLACET, line 83:
```suggestion
         * is transferred to the array. The user should not use \c array nor \c schema
```

include/sparrow/array_api.hpp
```diff
@@ -33,28 +43,96 @@ namespace sparrow
...
+        /**
+         * Constructs an empty array.
+         */
         SPARROW_API array() = default;
 
+        /**
+         * Constructs an array from the given typed layout. The ownership
+         * of the layout is transferred to the array.
+         *
+         * @param a An rvalue refernece to the typed layout.
```
Alex-PLACET, line 55:
```suggestion
         * @param a An rvalue reference to the typed layout.
```

Alex-PLACET (commented):
Some typos

.github/workflows/linux.yml
Alex-PLACET: I mean the setup of the CI will be faster

include/sparrow/layout/primitive_array.hpp
```diff
@@ -55,6 +53,21 @@ namespace sparrow
         using iterator_tag = std::random_access_iterator_tag;
     };
 
+    /**
+     * Array of values having the same binary size.
+     *
+     * The \c primitive_array class represents an array of values having the same binary size.
```
Klaim, line 59: Why the repetition?

include/sparrow/layout/primitive_array.hpp
```diff
@@ -55,6 +53,21 @@ namespace sparrow
...
+     * The current implementation supports types whose size is known at compile only.
```
Klaim, line 60:
```suggestion
     * The current implementation supports types whose size is known at compile time only.
```

include/sparrow/layout/primitive_array.hpp
```diff
@@ -55,6 +53,21 @@ namespace sparrow
...
+     *
+     * As the other arrays in sparrow, \c primitive_array<T> provides an API as if it was holding
+     * \c nullable<T> values instead of \c T values.
+     *
+     * Internally, the array contains a validity bitmap and a contiguous memory buffer
+     * holding the values.
+     *
+     * @tparam T the type of the values in the array.
+     * @see https://arrow.apache.org/docs/dev/format/Columnar.html#fixed-size-primitive-layout
```
Klaim, line 69: Move this line in the description?

include/sparrow/layout/primitive_array.hpp
```diff
@@ -55,6 +53,21 @@ namespace sparrow
         using iterator_tag = std::random_access_iterator_tag;
     };
 
+    /**
+     * Array of values having the same binary size.
```
Klaim, line 57: So there can be multiple types as long as they have the same binary size?

include/sparrow/layout/primitive_array.hpp
```diff
@@ -90,15 +103,43 @@ namespace sparrow
...
 
+        /**
+         * Constructs a primitive array with the passed range of values and an optional bitmap.
+         *
+         * The first argument can be any range of values as long as its value type is convertible
+         * to \c T.
+         * The second argument can be:
+         * - a bitmap range, i.e. a range of boolean-like values indicating the non-missing values.
+         *   The bitmap range and the value range must have the same size.
+         * \code{.cpp}
```
Klaim, line 114: Markdown syntax works too and is more readable in raw text.

Klaim (approved):
Overall good starting point, some tweaks could improve as suggested.

include/sparrow/layout/array_bitmap_base.hpp
```diff
@@ -19,9 +19,14 @@
 
 namespace sparrow
 {
-    /*
+    /**
      * Base class for arrays using a validity buffer for
      * defining their bitmap.
+     *
+     * @tparam D The derived type, i.e. the inheriting class for which
```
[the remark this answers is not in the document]
JohanMabille: Because at the point where we would put the reqiurement on D, the base has not be defined yet. Or I don't see which kind of requirement you're talking about.

include/sparrow/layout/primitive_array.hpp
JohanMabille: This is usually easier to find reference at the end in a dedicated function rather than drown in the middle of the description. Even if the description is quite short for now, it may grow.

include/sparrow/layout/primitive_array.hpp
JohanMabille: No, they all have the same type.

include/sparrow/layout/primitive_array.hpp
Klaim: I meant at the end of the description but before the section about the parameters (which will always be apart anyway in doxygen generated doc)

include/sparrow/layout/array_bitmap_base.hpp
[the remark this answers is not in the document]
Klaim: Ah ok `array_bitmap_base_impl` is defined below. It could be declared first and then required here. the requirement is checked at instantiation so that could work.

PR #265, opened 2024-11-02
Using PUBLIC_HEADER in cmake has the nasty property to flatten the di…

### LINKED ISSUE #263: Sparrow pollutes include dir
Running `make install` with a prefix set to `/tmp` does not only add files into `/tmp/include/sparrow` but also in `/tmp/include` which is unexpected and probably not desired

serge-sans-paille:
…rectory structure

Fix #263

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,14 +222,12 @@ if (UNIX)
     set_target_properties(
         sparrow
         PROPERTIES
-        PUBLIC_HEADER "${SPARROW_HEADERS}"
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
note: I'm not very happy with that change, it's just to exhibit the reason of the bug, I leave it to @JohanMabille or @Klaim to find a more adequate patch ;-)

JohanMabille:
Unfortunately I don't think we have a better option. Thanks for the fix!

PR #266, opened 2024-11-04
Normalize project versioning and soname

serge-sans-paille:
Instead of custom variable names, use the standard CMAKE_PROJECT_VERSION, PROJECT_VERSION_MAJOR etc.

Use them to properly set SOVERSION.


Alex-PLACET (approved):
LGTM

CMakeLists.txt
```diff
@@ -216,8 +217,10 @@ endif()
 
 add_library(sparrow SHARED ${SPARROW_HEADERS} ${SPARROW_SRC})
 
+set_target_properties(sparrow PROPERTIES VERSION ${CMAKE_PROJECT_VERSION}
+                              SOVERSION ${PROJECT_VERSION_MAJOR})
```
JohanMabille, line 221: We need to introduce binary versionning (which is different from API versionning) and set the so version accordingly.

SylvainCorlay:
@JohanMabille let's introduce what we did for soname with xeus.

SylvainCorlay (approved):

PR #271, opened 2024-11-06
Constructors run_end_encoded

DerThorsten:


src/run_encoded_array.cpp
```diff
@@ -49,4 +50,101 @@ namespace sparrow
...
+        // visit the acc_lengths array
+        std::int64_t length = 0;
+        std::int64_t null_count = 0;
+        acc_lengths_arr.visit([&](const auto& acc_lengths_array)
+            {
+                if constexpr(usable_array<std::decay_t<decltype(acc_lengths_array)>>)
+                {   
+                    auto acc_length_data = acc_lengths_array.data();
+                    // get the length of the array (ie last element in the acc_lengths array)
+                    auto acc_lengths = acc_lengths_array.data();
```
Alex-PLACET, line 73: Is it a duplicate of l.71 ? I don't see where is it used

DerThorsten: good catch!Not sure why it was not reported as unused variable

JohanMabille (approved):

Alex-PLACET (approved):

PR #274, opened 2024-11-07
Change minimum requirements for OSX

Alex-PLACET:


.github/workflows/osx.yml
```diff
@@ -18,8 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - 13
-        - 14
```
JohanMabille, line 22: I think we want to exclude OSX13 - AppleClang and OSX14 - AppleClang, but keep OSX13 - clang and OSX14 - clang

Alex-PLACET: Fixed 
image

JohanMabille (approved):

PR #277, opened 2024-11-08
Add growth factor to buffer

Alex-PLACET:


include/sparrow/buffer/buffer.hpp
```diff
@@ -726,6 +732,15 @@ namespace sparrow
         }
     }
 
+    template<class T>
+    constexpr void buffer<T>::reserve_with_growth_factor(size_type new_cap)
+    {
+        if(new_cap > capacity())
+        {
+            reserve(new_cap * SPARROW_BUFFER_GROWTH_FACTOR);
+        }
```
JohanMabille, line 741: I think this logis should actually be in `reserve` since it's a public API. And `reserve` shoudl forward to a private method (`reserve_impl`?) when needed.

JohanMabille (changes requested):
I think the "reserve workflow" private => public shoud be done the opposite way.

include/sparrow/buffer/buffer.hpp
Alex-PLACET: I checked the behavior of std::vector. The reserve, will change to the exact requested capacity (if the new cap is bigger than the current one).
With the insert and push_back, the capacity reserved will be bigger than the size if the new size is bigger than the current capacity.
So my current implementation reproduce the same

JohanMabille: aaah yes my bad!

JohanMabille (approved):

PR #283, opened 2024-11-13
variable sized binary constructors and minor refactor

DerThorsten:
* added convenience constructors
*  renamed variable_sized_binary to variable_sized_binary_impl
* create typedef to the actual arrays without templates
* added helper to use `range_size` instead of `std::ranges::size` because that only works for sized ranges
   (and the use of  join_view in the convenience constructors yields a non-sized range)

needs #282

### DIFF
--- a/include/sparrow/layout/variable_size_binary_array.hpp
+++ b/include/sparrow/layout/variable_size_binary_array.hpp
@@ -20,21 +20,67 @@
 #include <numeric>
 #include <ranges>
 #include <string>
+#include <vector>
+#include <cstddef>
 
+#include "sparrow/arrow_interface/arrow_array.hpp"
+#include "sparrow/arrow_interface/arrow_schema.hpp"
 #include "sparrow/arrow_array_schema_proxy.hpp"
 #include "sparrow/buffer/buffer_adaptor.hpp"
+#include "sparrow/buffer/u8_buffer.hpp"
 #include "sparrow/layout/array_bitmap_base.hpp"
 #include "sparrow/layout/layout_iterator.hpp"
 #include "sparrow/types/data_type.hpp"
 #include "sparrow/utils/contracts.hpp"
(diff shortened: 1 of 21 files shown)


include/sparrow/layout/variable_size_binary_array.hpp
```diff
@@ -20,31 +20,77 @@
...
+        concept char_like = std::same_as<T, char> || std::same_as<T, std::byte> || std::same_as<T, std::uint8_t>;
+    }    
+
+
+
+    template <std::ranges::sized_range T, class CR, layout_offset OT>
+    class variable_size_binary_array_impl;
+
+    using string_array =     variable_size_binary_array_impl<std::string, std::string_view, std::int32_t>;
+    using big_string_array = variable_size_binary_array_impl<std::string, std::string_view, std::int64_t>;
```
Alex-PLACET, line 82: We should also define binary_array and big_binary_array ? In a future PR ?

Alex-PLACET (approved):

include/sparrow/utils/ranges.hpp
```diff
@@ -0,0 +1,34 @@
...
+    template<std::ranges::input_range R>
+    requires(std::ranges::sized_range<R>)
+    auto range_size(R && r)
+    {
+        return std::ranges::size(r);
+    }
+
+    template<std::ranges::input_range R>
+    requires(!std::ranges::sized_range<R>)
+    auto range_size(R && r)
```
JohanMabille, line 29: I think we should constraint the return type to be size_t: this way, both overlaods have the same return type and the user doe snot have to static cast

include/sparrow/layout/variable_size_binary_array.hpp
DerThorsten: I felt like this was better for a follow up pr

JohanMabille (approved):

PR #287, opened 2024-11-14
Add slice method to array

Alex-PLACET:


src/array.cpp
```diff
@@ -48,6 +49,13 @@ namespace sparrow
         return array_element(*p_array, index);
     }
 
+    void array::slice(size_type start, size_type end)
```
DerThorsten, line 52: I would prefer slice to return a array which is a view to the original array.
ie auto sliced_arr = arr.slice(start, end);
With the view function of the arrow_proxy it should be possible to create a cheap "view-copy" of the array and than change the offset and size there

test/test_array.cpp
```diff
@@ -251,10 +256,54 @@ namespace sparrow
...
+            REQUIRE_EQ(ar.size(), 4);
+            CHECK_EQ(std::get<const_reference>(ar[0]), make_nullable<scalar_value_type>(1));
+            CHECK_EQ(std::get<const_reference>(ar[1]), make_nullable<scalar_value_type>(2));
+            CHECK_EQ(std::get<const_reference>(ar[2]), make_nullable<scalar_value_type>(3));
+            CHECK_EQ(std::get<const_reference>(ar[3]), make_nullable<scalar_value_type>(4));
+
+            ar.slice(2, 8);
+
+            REQUIRE_EQ(ar.size(), 6);
+            CHECK_EQ(std::get<const_reference>(ar[0]), make_nullable<scalar_value_type>(2));
```
DerThorsten, line 300: we should add this to the consistency test instead of implementing this by hand

src/array.cpp
Alex-PLACET: Now we have arr.slice and arr.slice_view

src/array.cpp
```diff
@@ -77,6 +77,29 @@ namespace sparrow
...
+    array array::slice_view(size_type start, size_type end) const
+    {
+        SPARROW_ASSERT_TRUE(start <= end);
+        const arrow_proxy& arrow_proxy_copy = get_arrow_proxy();
+        ArrowSchema* as = new ArrowSchema;
+        *as = arrow_proxy_copy.schema();
+        ArrowArray* ar = new ArrowArray;
+        *ar = arrow_proxy_copy.array();
+        ar->offset = static_cast<int64_t>(start);
+        ar->length = static_cast<int64_t>(end - start);
```
JohanMabille, line 99: This method creates a memory leak, since `as` and `ar` will not be deleted. A way to fix this would be to allocate these structures on the stack and move them in the returned array. The release method should be set to empty methods to avoid destruciton of the data when the view is destroyed.

JohanMabille (changes requested):

src/array.cpp
Alex-PLACET: You are right. But we can't set the release to null as it means that the array is invalid.
I have to create another release callback which don't delete the buffers

JohanMabille (approved):

src/array.cpp
```diff
@@ -77,6 +79,29 @@ namespace sparrow
...
+        arrow_proxy& arrow_proxy_copy = copy.get_arrow_proxy();
+        arrow_proxy_copy.set_offset(start);
+        arrow_proxy_copy.set_length(end - start);
+        return copy;
+    }
+
+    array array::slice_view(size_type start, size_type end) const
+    {
+        SPARROW_ASSERT_TRUE(start <= end);
+        const arrow_proxy& arrow_proxy_copy = get_arrow_proxy();
```
DerThorsten, line 95: why not 
```C++
auto view_proxy = get_arrow_proxy().view();
view_proxy.set_offset(start);
```

Alex-PLACET: View will use existing array and schema, it just use pointers to the c structures. In my case I have to copy the C structures to modify the offset and length withour changing the original and but I keep the pointers values of the members

DerThorsten (approved):

PR #288, opened 2024-11-15
Fix the use of clang with VS build system when required

Alex-PLACET:


test/better_junit_reporter.hpp
```diff
@@ -136,7 +136,7 @@ void JUnitReporter::test_run_start()
     }
 }
 
-void print_failures(const sparrow::JUnitTestSuites& test_suites)
+static void print_failures(const sparrow::JUnitTestSuites& test_suites)
```
JohanMabille, line 139: `inline` would be better than `static`

JohanMabille (approved):

test/better_junit_reporter.hpp
JohanMabille, line 139:
```suggestion
inline void print_failures(const sparrow::JUnitTestSuites& test_suites)
```

DerThorsten (approved):

PR #289, opened 2024-11-18
More doc and missing APIs

JohanMabille:


docs/source/typed_array.md
```diff
@@ -0,0 +1,227 @@
...
+
+sp::nullable<int> n = 2;
+std::cout << n.has_value() << std::endl; // Prints true
+std::cout << n.value() << std::endl;     // Prints 2
+
+stp::nullable<double> nd = sp::nullval;
+std::cout << nd.has_value() << std::end; // Prints false
+```
+
+Conrary to [std::optional](https://en.cppreference.com/w/cpp/utility/optional), \ref nullable
```
Alex-PLACET, line 32:
```suggestion
Contrary to [std::optional](https://en.cppreference.com/w/cpp/utility/optional), \ref nullable
```

include/sparrow/layout/array_base.hpp
```diff
@@ -163,6 +183,25 @@ namespace sparrow
...
+     * in the array with bounds checking.
+     * @param i the index of the element in the array.
+     * @throw std::out_of_range if \c i is not within the range of the container.
+     */
+    template <class D>
+    auto array_crtp_base<D>::at(size_type i) const -> const_reference
+    {
+        if (i >= size())
+        {
+            std::ostringstream oss117;
```
Alex-PLACET, line 197: BonjourHiGIF

test/test_primitive_array.cpp
```diff
@@ -54,11 +54,34 @@ namespace sparrow
...
+            {
+                CHECK_FALSE(ar.empty());
+            }
+
             SUBCASE("constructor")
             {
                 CHECK_EQ(ar.size(), 4);
             }
 
+            SUBCASE("at")
```
Alex-PLACET, line 67: These new tests should be added to all layouts, right ?

include/sparrow/layout/array_base.hpp
```diff
@@ -178,6 +217,26 @@ namespace sparrow
         );
     }
 
+    /**
+     * Returns a constant reference to the first element in the container.
```
Alex-PLACET, line 221: Can you add information about the contract: should not be empty

include/sparrow/layout/array_base.hpp
```diff
@@ -178,6 +217,26 @@ namespace sparrow
...
+     */
+    template <class D>
+    auto array_crtp_base<D>::front() const -> const_reference
+    {
+        SPARROW_ASSERT_TRUE(!empty());
+        return (*this)[size_type(0)];
+    }
+
+    /**
+     * Returns a constant reference to the last element in the container.
```
Alex-PLACET, line 231: Can you add information about the contract: should not be empty

test/test_primitive_array.cpp
JohanMabille: Yes, but I would prefer to do it in a dedicated PR to avoid making this one too heavy. Notice that since these new functions all rely on `operator[]` that is tested for all layouts, we know that they work for all layouts (for now, I'm not saying that we should not add tests).

Alex-PLACET (approved):

PR #294, opened 2024-11-21
Formatters for sparrow classes

Alex-PLACET:


include/sparrow/layout/list_layout/list_value.hpp
```diff
@@ -45,3 +45,26 @@ namespace sparrow
...
+    constexpr auto parse(std::format_parse_context& ctx)
+    {
+        return ctx.begin();  // Simple implementation
+    }
+
+    auto format(const sparrow::list_value& list_value, std::format_context& ctx) const
+    {
+
+        return std::format_to(ctx.out(), "[size={}] TODO", list_value.size());
+        // for (std::size_t i = 0; i < list_value.size(); ++i)
```
DerThorsten, line 62: to remove

include/sparrow/array_api.hpp
```diff
@@ -14,6 +14,9 @@
 
 #pragma once
 
+#include <cctype>
+#include <limits>
```
JohanMabille, line 18: Why these additional includes?

include/sparrow/arrow_array_schema_proxy.hpp
```diff
@@ -480,5 +484,113 @@ namespace sparrow
...
+#if defined(__cpp_lib_format)
+
+template <>
+struct std::formatter<sparrow::buffer_view<uint8_t>>
+{
+private:
+
+    char delimiter = ' ';
+    std::string_view opening = "[";
+    std::string_view closing = "]";
```
JohanMabille, line 498: Should these variable be `constexpr` ?

include/sparrow/layout/array_base.hpp
```diff
@@ -445,6 +463,82 @@ namespace sparrow
         return std::ranges::equal(lhs, rhs);
     }
 
-    
+    constexpr unsigned long stoul(const std::string_view str)
```
JohanMabille, line 466: I may have missed it, but I don't see this function used. If so, I think this should be moved in a util file instead. Also this won't work with ADL (but I guess we don't have the choice, since it is forbidden to overload in the standard namespace).

include/sparrow/layout/array_base.hpp
```diff
@@ -15,16 +15,17 @@
 #pragma once
 
 #include <algorithm>
+#include <concepts>
```
JohanMabille, line 18: Why adding this header here?

include/sparrow/layout/array_base.hpp
```diff
@@ -127,6 +128,9 @@ namespace sparrow
         const_bitmap_range bitmap() const;
         const_value_range values() const;
 
+        [[nodiscard]] std::optional<const std::string_view> name() const;
+        [[nodiscard]] std::optional<const std::string_view> metadata() const;
```
JohanMabille, line 132: I think `std::optional<std:string_view>` would be enough since `std::string_view` only provides a const API.

include/sparrow/utils/format.hpp
```diff
@@ -0,0 +1,204 @@
...
+#include <format>
+#include <numeric>
+#include <ranges>
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "sparrow/utils/contracts.hpp"
+
+#pragma once
```
JohanMabille, line 26: Should be before the include

include/sparrow/utils/format.hpp
```diff
@@ -0,0 +1,204 @@
...
+                SPARROW_ASSERT_TRUE(std::ranges::size(columns[0]) == std::ranges::size(*it));
+            }
+        }
+
+        std::vector<size_t> widths = columns_widths(columns);
+        
+        // max with names
+        for (size_t i = 0; i < std::ranges::size(headers); ++i)
+        {
+            #if defined(__clang__)
```
JohanMabille, line 153: Would a `static_cast` avoid the warning here? I think it would be easier to read. If not, then maybe a single set of pragma enclosing the whole function would probably improve the readability.

include/sparrow/utils/nullable.hpp
```diff
@@ -146,65 +156,60 @@ namespace sparrow
...
         template <class To1, class From1, class To2, class From2>
-        concept both_assignable_from_cref =
-            std::is_assignable_v<std::add_lvalue_reference_t<To1>, mpl::add_const_lvalue_reference_t<From1>> and
-            std::is_assignable_v<std::add_lvalue_reference_t<To2>, mpl::add_const_lvalue_reference_t<From2>>;
+        concept both_assignable_from_cref = std::is_assignable_v<
+                                                std::add_lvalue_reference_t<To1>,
+                                                mpl::add_const_lvalue_reference_t<From1>>
+                                            and std::is_assignable_v<
+                                                std::add_lvalue_reference_t<To2>,
+                                                mpl::add_const_lvalue_reference_t<From2>>;
```
JohanMabille, line 204: I find this version harder to read than the original one (but that should be fixed in the formatting config, not in that PR).

include/sparrow/array_api.hpp
Alex-PLACET: Hmm, maybe about some debugging at some point. I will remove it

include/sparrow/layout/array_base.hpp
Alex-PLACET: At some point, I was working on the "parse" part to be able to change the way we format the variable. I dropped that for the moment.
Let's delete this unused function

include/sparrow/utils/format.hpp
Alex-PLACET: No because, as headers is a template argument, depend ending of its type, `std::ranges::size(headers)` will not return the same type, in one case we have teh warning about the sign conversion, in the other a warning about useless static_cast.

include/sparrow/layout/array_base.hpp
Alex-PLACET: done

include/sparrow/utils/format.hpp
Alex-PLACET: done

include/sparrow/layout/array_base.hpp
Alex-PLACET: fixed

include/sparrow/utils/format.hpp
JohanMabille: I still think a single set of pragma enclosing the whole function would make it easier to read.

include/sparrow/layout/list_layout/list_array.hpp
```diff
@@ -735,3 +735,16 @@ namespace sparrow
...
+// template <typename U>
+// struct std::formatter<sparrow::list_array, U> : std::formatter<sparrow::array, U>
+// {
+//     auto format(const sparrow::list_array& arr, U& ctx) -> decltype(ctx.out())
+//     {
+//         return std::formatter<sparrow::array, U>::format(arr, ctx);
+//     }
+// };
+
+// #endif
```
JohanMabille, line 750: Is that intentional or is it a leftover?

JohanMabille (approved):
LGTM once the two remaining comments are addressed

include/sparrow/layout/list_layout/list_array.hpp
Alex-PLACET: removed

include/sparrow/utils/format.hpp
Alex-PLACET: done

PR #300, opened 2024-11-27
Binviewlayout wip + refactor

DerThorsten:


src/arrow_interface/arrow_array.cpp
```diff
@@ -40,31 +40,131 @@ namespace sparrow
...
+        {
+            return 16;
+        }
+        else
+        {
+            // get the position of second comma
+            const auto second_comma = std::strchr(format, ',') + 1;
+            const auto num_bits = std::atoi(second_comma);
+            
+            SPARROW_ASSERT_TRUE(num_bits == 32 || num_bits == 64 || num_bits == 128 || num_bits == 256);
```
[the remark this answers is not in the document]
DerThorsten: I am not sure if 8,16 would be allowed. arrow-cpp only defines types for the values above

Alex-PLACET (approved):

include/sparrow/layout/variable_size_binary_view_array.hpp
```diff
@@ -0,0 +1,369 @@
...
+#include "sparrow/utils/ranges.hpp"
+
+namespace sparrow
+{
+
+
+    template <class T>
+    class variable_size_binary_view_array_impl;
+    using string_view_array = variable_size_binary_view_array_impl<std::string_view>;
+    using binary_view_array = variable_size_binary_view_array_impl<std::span<const std::byte>>;
```
JohanMabille, line 41: Since `std::span` is a const view, I don't think it is necessary to constify its value type. I would prefer that we use buffer_view instead, since `std::span` implementation is very incomplete (no comparison operators, no cbegin/cend methods, etc...), but I struggle to get it working with the variable_size_binary_array.

include/sparrow/layout/variable_size_binary_view_array.hpp
```diff
@@ -0,0 +1,369 @@
...
+    };
+
+    /**
+     * Checks whether T is a variable_size_binary_view_array_impl type.
+     */
+    template <class T>
+    constexpr bool is_variable_size_binary_view_array_impl_v = is_variable_size_binary_view_array_impl<T>::value;
+
+    template <class T>
+    class variable_size_binary_view_array_impl final : public mutable_array_bitmap_base<variable_size_binary_view_array_impl<T>>
```
JohanMabille, line 97: Why not `variable_size_binary_view_array` instead? (I am doing the same in variable_size_binary_array btw).

DerThorsten: the users should not use use class with templates because only where few instantations are allowed

JohanMabille: Indeed but they will use the "is_variable_size_binary_view_array_impl", so this is not consistent. Or we provide `is_string_view_array` and  `is_binary_view_array` that compare to the instantiations instead of having a generic comparison.

JohanMabille (approved):

PR #305, opened 2024-12-02
Add nanoarrow dependency and a test with primitive array

Alex-PLACET:


Alex-PLACET:
@paleolimbot I was not aware of ArrowArrayViewValidate, thank you !
I created an issue to add Sparrow to archery, but we are not ready yet. 👍

.github/workflows/docs.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
         cache-downloads: true
 
     - name: Configure using CMake
-      run: cmake -G Ninja -Bbuild -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DBUILD_TESTS=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_DOCS=ON
```
Alex-PLACET, line 24: Useless

.github/workflows/windows.yml
```diff
@@ -136,11 +143,44 @@ jobs:
...
+        -DCMAKE_INSTALL_PREFIX=$SPARROW_INSTALL_PREFIX \
+        -DBUILD_TESTS=ON \
+        -DBUILD_EXAMPLES=ON \
+        -DUSE_DATE_POLYFILL=${{matrix.toolchain.date-polyfill}} \
+        $GENERATOR_EXTRA_FLAGS \
+        $SPARROW_CHECKS \
+        $SPARROW_ADDITIONAL_OPTIONS \
+        -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+        -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+        -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL
```
Alex-PLACET, line 183: We compile nanoarrow in release so we have to explicitly define CMAKE_MSVC_RUNTIME_LIBRARY to avoid binary mismatch

include/sparrow/buffer/dynamic_bitset/bitset_iterator.hpp
```diff
@@ -94,6 +94,10 @@ namespace sparrow
     {
         if constexpr (is_const)
         {
+            if(p_bitset->data() == nullptr)
```
Alex-PLACET, line 97: Nanoarrow don't create bitmap buffer if all elements are no null. But in our code we always create a dynamic_bitset, which can take as input a nullptr but until this fix, don't really support it.
In the case of const methods, I always return true.
For non const I assert. We should not reach this assert as we don't permit the modification of non sparrow arrays.

DerThorsten:
nitpick: with all the formating changes its hard to see what the actual changes are.
We should do the global formating at some point

JohanMabille (approved):

DerThorsten (approved):

PR #316, opened 2024-12-20
Remove useless algorithm

Alex-PLACET:


.github/workflows/linux.yml
```diff
@@ -105,9 +105,7 @@ jobs:
       if: success() || failure()
       with:
         name: test_sparrow_lib_report_Linux_${{ matrix.sys.compiler }}_${{ matrix.sys.version }}_${{ matrix.sys.stdlib }}_${{ matrix.config.name }}_date-polyfill_${{ matrix.sys.date-polyfill}}
-        path: |
-          build/test/test_sparrow_lib_report_doctest.xml
-          build/test/test_sparrow_lib_report_catch2.xml
+        path: ./**/test*.xml
```
Alex-PLACET, line 108: Because the only catch2 test is the one for algorithm.hpp, we deactivate it. And no use use wildcard to have less maintenance

JohanMabille (approved):

PR #323, opened 2025-01-10
Fix test_format.cpp

Alex-PLACET:


test/test_format.cpp
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <version>
```
Alex-PLACET, line 15: Without that, __cpp_lib_format is undefined

test/test_format.cpp
```diff
@@ -133,45 +135,6 @@ TEST_SUITE("format")
         }
     }
 
-    TEST_CASE("to_header")
```
Alex-PLACET, line 136: It was removed

JohanMabille (approved):

PR #324, opened 2025-01-20
Timestamp array

Alex-PLACET:


JohanMabille (approved):
The need to add friendship in `primitive_array` and to define `arrow_traits` for duration types because they are internally hold in a `primitive_array` while their associated layout will be another type is an argument for captuing out of `primitive_array` the common implementation. This should be done in a dedicated PR though.

PR #333, opened 2025-01-30
Add duration_array

Alex-PLACET:


include/sparrow/types/data_traits.hpp
```diff
@@ -116,31 +116,31 @@ namespace sparrow
     {
         static constexpr data_type type_id = data_type::DURATION_SECONDS;
         using value_type = std::chrono::seconds;
-        using const_reference = std::chrono::seconds;
+        using const_reference = const std::chrono::seconds&;
     };
```
JohanMabille, line 120: I think this traits and the following can inherit from `common_native_type_traits`

Alex-PLACET: ok done

JohanMabille (approved):

PR #335, opened 2025-01-31
Interval array

Alex-PLACET:


include/sparrow/layout/temporal/interval_types.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace sparrow
+{
+    namespace chrono
+    {
+        using days = std::chrono::duration<int32_t, std::ratio<86400>>;
+        using months = std::chrono::duration<int32_t, std::ratio<2629746>>;
+    }
+
+#pragma pack(push, 1)
```
JohanMabille, line 30: I think this deserves a comment explaining why this is required.

JohanMabille (approved):

PR #336, opened 2025-01-31
Added record bact import from/export to struct_array

JohanMabille:


include/sparrow/record_batch.hpp
```diff
@@ -169,6 +196,25 @@ namespace sparrow
...
+        std::transform(
+            m_array_list.cbegin(),
+            m_array_list.cend(),
+            m_name_list.begin(),
+            [](const array& ar)
+            {
+                return ar.name();
+            }
+        );
+        init_array_map();
```
Alex-PLACET, line 214: Can we move these logic to dedicated functions which returns the values to put in the members ?

src/array.cpp
```diff
@@ -47,6 +47,19 @@ namespace sparrow
         return p_array->data_type();
     }
 
+    std::string_view array::name() const
+    {
+        ArrowSchema* schema = get_arrow_schema(const_cast<array&>(*this));
+        std::string_view res((schema->name != nullptr) ? schema->name : "");
+        return res;
+    }
```
Alex-PLACET, line 55: I suggest to not set an empty string and return an optional, and btw use the proxy implementation.
```suggestion
    std::optional<std::string_view> array::name() const
    {
        return get_arrow_proxy().name();
    }
```

src/array.cpp
```diff
@@ -47,6 +47,19 @@ namespace sparrow
...
+
+    void array::set_name(std::string_view name)
+    {
+        arrow_proxy& proxy = detail::array_access::get_arrow_proxy(*this);
+        proxy.set_name(name);
+    }
```
Alex-PLACET, line 61:
```suggestion
    void array::set_name(std::optional<std::string_view> name)
    {
        get_arrow_proxy().set_name(name);
    }
```

include/sparrow/record_batch.hpp
JohanMabille: I'm not sure we want to do that, this would require an additional memory allocation instead of filling the members in plae. Besides, this transform is used only here. What is the issue with this implementation?

Alex-PLACET: IMO it's cleaner to have only a constructor initialization list.

``` c++

std::vector<name_type> get_names(const std::vector<array>& array_list)
{
    const auto names = array_list | std::views::transform(
                    [](const array& ar)
                    {
                        return ar.name();
                    }
    return std::vector<name_type> {names.begin(), names.end()};
}

  template <std::ranges::input_range CR>
        requires std::same_as<std::ranges::range_value_t<CR>, array>
    record_batch::record_batch(CR&& columns)
        : m_array_list(to_vector<array>(std::move(columns)))
        , m_name_list(get_names(m_array_list))
        , 
    {
        init_array_map(); 
```

Alex-PLACET (approved):

include/sparrow/record_batch.hpp
JohanMabille: Oh indeed, I did not get that.

PR #357, opened 2025-02-17
Add documentation and new constructor (pointer + size) for u8_buffer,…

Alex-PLACET:
… add operator[] too

### DIFF
--- a/include/sparrow/buffer/allocator.hpp
+++ b/include/sparrow/buffer/allocator.hpp
@@ -237,6 +237,7 @@ namespace sparrow
 #if defined(__GNUC__)
 #    pragma GCC diagnostic push
 #    pragma GCC diagnostic ignored "-Wmismatched-new-delete"
+#    pragma GCC diagnostic ignored "-Wnull-dereference"
 #endif
                 return allocator.deallocate(p, n);
 #if defined(__GNUC__)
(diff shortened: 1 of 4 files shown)


include/sparrow/buffer/allocator.hpp
```diff
@@ -111,12 +111,19 @@ namespace sparrow
 
             [[nodiscard]] T* allocate(std::size_t n) override
             {
-                return m_alloc.allocate(n);
+                return reinterpret_cast<T*>(m_alloc.allocate(n));
```
JohanMabille, line 114: This `reinterpret_cast` should not be required, we should ensure that `T` and `A::value_type` are the same in the `any_allocator` constructor. This would also allow to remove the reinterpret cast in the deallocate, and probably remove the surrounding pragma.

I understand this is required so that the current implementation of the u8_buffer can take owneship of an already allocated buffer of a different type, and I don't want to delay this too much, but I think this should be refactored with a specific allocator adaptor in the near future (so maybe just add a TODO when you restore the GCC pragmas).

JohanMabille (approved):

include/sparrow/buffer/u8_buffer.hpp
```diff
@@ -72,17 +78,43 @@ namespace sparrow
         using holder_type = detail::holder<buffer<std::uint8_t>>;
         using buffer_adaptor_type = buffer_adaptor<T, buffer<std::uint8_t>&>;
         using holder_type::extract_storage;
+        using buffer_adaptor_type::operator[];
```
JohanMabille, line 81: This should not be required since the inheritance from `buffer_adaptor` is public. Is that a workaround for a compiler issue?

JohanMabille (approved):

PR #358, opened 2025-02-18
Add cmake option to compile library in shared or static

Alex-PLACET:


JohanMabille (commented):
The name of the static library and dynamic library on Windows should be different, otherwise it is impossible to build both of them at the same time.

Alex-PLACET:
@JohanMabille https://github.com/man-group/sparrow/issues/359

SylvainCorlay:
I think that `SPARROW_BUILD_SHARED` and `SPARROW_BUILD_STATIC` should be settable independently.

PR #360, opened 2025-02-25
Fix GCC compilation on Windows

Alex-PLACET:


CMakeLists.txt
```diff
@@ -320,7 +320,13 @@ if(ENABLE_COVERAGE)
     enable_coverage(sparrow)
 endif()
 
-if(BUILD_TESTS)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
JohanMabille, line 323: This should be done on Windows only.

JohanMabille (changes requested):

JohanMabille (approved):

PR #370, opened 2025-03-03
Set default default value for USE_LARGE_INT_PLACEHOLDERS option

Alex-PLACET:


JohanMabille (commented):
Should we removed the settings of USE_LARGE_INT_PLACEHOLDERS in [Windows workflow](https://github.com/man-group/sparrow/blob/main/.github/workflows/windows.yml#L22-L24) to test the default setting is correclty applied?

PR #371, opened 2025-03-04
Better metadata

Alex-PLACET:


include/sparrow/layout/null_array.hpp
```diff
@@ -192,6 +198,38 @@ namespace sparrow
     {
         return m_index < rhs.m_index;
     }
+
+    template <input_metadata_container METADATA_RANGE>
+    arrow_proxy
+    null_array::create_proxy(size_t length, std::optional<std::string_view> name, std::optional<METADATA_RANGE> metadata)
```
Alex-PLACET, line 204: moved from cpp

include/sparrow/layout/run_end_encoded_layout/run_end_encoded_array.hpp
```diff
@@ -127,8 +128,61 @@ namespace sparrow
 
     SPARROW_API
     bool operator==(const run_end_encoded_array& lhs, const run_end_encoded_array& rhs);
+
+    template <input_metadata_container METADATA_RANGE>
+    auto run_end_encoded_array::create_proxy(
```
Alex-PLACET, line 133: moved from cpp

include/sparrow/layout/union_array.hpp
```diff
@@ -534,6 +538,74 @@ namespace sparrow
         );
     }
 
+    template <input_metadata_container METADATA_RANGE>
+    auto dense_union_array::create_proxy_impl(
```
Alex-PLACET, line 542: moved from cpp

include/sparrow/utils/metadata.hpp
```diff
@@ -0,0 +1,196 @@
...
+        {
+            return !(lhs == rhs);
+        }
+
+    private:
+
+        SPARROW_API std::string_view extract_string_view();
+        SPARROW_API void extract_key_value();
+
+        const key_value_view& m_parent;
```
JohanMabille, line 78: Storing a pointer default initialized to nullptr would allow the generation of copy and move semantic (and you can even make the iterator default constructible. I think the `std::ranges::view_interface` make ssome assumptions about the iterator used in the inheriting class (I remember being hit by it for the repeat_view).

include/sparrow/utils/metadata.hpp
```diff
@@ -0,0 +1,196 @@
...
+        int32_t m_index;
+        const char* m_current;
+        std::string_view m_key;
+        std::string_view m_value;
+    };
+
+    // Custom view to lazily extract key/value pairs from the buffer
+    class key_value_view : public std::ranges::view_interface<key_value_view>
```
JohanMabille, line 86: I guess this is the reason for dropping GCC12 support? If so, we can try to reintroduce it after this PR is merged.

include/sparrow/utils/metadata.hpp
```diff
@@ -0,0 +1,196 @@
...
+        int32_t m_num_pairs = 0;
+
+        friend key_value_view_iterator;
+    };
+
+    template <typename T>
+    concept input_metadata_container = std::ranges::input_range<T>
+                                       && std::same_as<std::ranges::range_value_t<T>, metadata_pair>;
+
+    std::string get_metadata_from_key_values(const input_metadata_container auto& metadata)
```
JohanMabille, line 114: Styling: an template method would be more consistent with the existing code base rather than an auto function.

src/arrow_array_schema_proxy.cpp
```diff
@@ -301,24 +301,13 @@ namespace sparrow
...
+    [[nodiscard]] std::optional<key_value_view> arrow_proxy::metadata() const
     {
         if (schema().metadata == nullptr)
         {
             return std::nullopt;
         }
-        return std::string_view(schema().metadata);
-    }
-
-    void arrow_proxy::set_metadata(std::optional<std::string_view> metadata)
```
JohanMabille, line 313: Was this method indirectly exposed ot the end user? If so, how can a user set the metadata now?

JohanMabille (changes requested):

src/arrow_array_schema_proxy.cpp
Alex-PLACET: This function is now templated and so moved to the header

include/sparrow/utils/metadata.hpp
Alex-PLACET: ok

include/sparrow/utils/metadata.hpp
Alex-PLACET: Fixed

JohanMabille (approved):

PR #373, opened 2025-03-06
Implemented primitive_data_access specialization for bool

JohanMabille:


include/sparrow/buffer/dynamic_bitset/dynamic_bitset.hpp
```diff
@@ -98,18 +100,22 @@ namespace sparrow
...
     {
+        base_type::zero_unused_bits();
     }
 
     template <std::integral T>
     constexpr dynamic_bitset<T>::dynamic_bitset(block_type* p, size_type n, size_type null_count)
         : base_type(storage_type(p, this->compute_block_count(n)), n, null_count)
     {
+        base_type::zero_unused_bits();
+        SPARROW_ASSERT_TRUE(base_type::null_count() == base_type::size() - base_type::count_non_null());
```
Alex-PLACET, line 118: Should we really have this assert or should we have this in unit test?

include/sparrow/layout/primitive_layout/primitive_data_access.hpp
```diff
@@ -278,5 +355,175 @@ namespace sparrow
...
+            , m_view(get_data_view())
+            , m_dummy_buffer()
+            , m_adaptor(&m_dummy_buffer, 0u)
+            , m_empty_adaptor(true)
+        {
+            if (p_proxy->is_created_with_sparrow())
+            {
+                m_adaptor = get_data_adaptor();
+                m_empty_adaptor = false;
+            }
```
Alex-PLACET, line 375: Can you set the m_adaptor and m_empty_adaptor thanks static functions

bitset_adaptor bitset_adaptor_init(const arrow_proxy& proxy)
{
if (p_proxy->is_created_with_sparrow())
            {

include/sparrow/buffer/dynamic_bitset/dynamic_bitset.hpp
JohanMabille: I think we must keep it to detect when the user creates a bitset in an incorrect state.

PR #385, opened 2025-04-01
Add gcc 12 linux ci

Alex-PLACET:


include/sparrow/utils/ranges.hpp
```diff
@@ -69,6 +69,38 @@ namespace sparrow
...
+            template <std::ranges::input_range R, std::weakly_incrementable O>
+                requires std::indirectly_copyable<std::ranges::iterator_t<R>, O>
+            constexpr std::ranges::copy_result<std::ranges::borrowed_iterator_t<R>, O>
+            operator()(R&& r, O result) const
+            {
+                return (*this)(std::ranges::begin(r), std::ranges::end(r), std::move(result));
+            }
+        };
+
+        inline constexpr copy_fn copy;
```
JohanMabille, line 102: I think we should have a ` copy` function that calls ` copy_fn`  only when ` std::ranges::copy` is no available, otherwise we shoulc forward to tthe standard implementation.

PR #386, opened 2025-04-01
Add gcc 11 linux ci

Alex-PLACET:


include/sparrow/utils/constexpr.hpp
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+namespace sparrow
+{
+// If using gcc versionn < 12, we define the constexpr keyword to be empty.
+#if defined(__GNUC__) && __GNUC__ < 12
+#    define SPARROW_CONSTEXPR_GCC_11 inline
+#else
+#    define SPARROW_CONSTEXPR_GCC_11 constexpr
+#endif
```
JohanMabille, line 22: This should be moved to config.hpp, I find it overkill to have a file just for this definition.

include/sparrow/utils/ranges.hpp
```diff
@@ -92,7 +92,14 @@ namespace sparrow
             {
                 for (; first != last; ++first, (void) ++result)
                 {
+#if defined(__GNUC__)
```
JohanMabille, line 95: let's add ` && __GNUC__ < 12`  so we know that we can remove this when we drop support for gcc11

include/sparrow/utils/ranges.hpp
```diff
@@ -92,7 +92,14 @@ namespace sparrow
...
+#    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wnull-dereference"
+#endif
                     *result = *first;
+#if defined(__GNUC__)
```
JohanMabille, line 100: Same here

JohanMabille (changes requested):

include/sparrow/utils/ranges.hpp
Alex-PLACET: ok

include/sparrow/utils/ranges.hpp
Alex-PLACET: ok

include/sparrow/utils/constexpr.hpp
Alex-PLACET: ok

JohanMabille (approved):

PR #400, opened 2025-04-23
Bitmap nullptr support

Alex-PLACET:


include/sparrow/buffer/buffer_view.hpp
```diff
@@ -342,7 +344,14 @@ namespace sparrow
...
     buffer_view<T>::operator buffer<std::remove_const_t<T>>() const
     {
-        return {p_data, p_data + m_size};
+        if (!p_data)
+        {
+            return buffer<std::remove_const_t<T>>();
+        }
+        else
+        {
+            return buffer<std::remove_const_t<T>>(cbegin(), cend());
```
JohanMabille, line 353: Why not returning this all the time? The buffer constructor should handle the case where ` cbegin() == cend()`. If not it should be fixed.

Alex-PLACET: The constructor which takes iterators does an allocation.

JohanMabille: Yes but since cbegin() == cend(), the allocation should be 0 byte, i.e. a noop.

JohanMabille (approved):

PR #401, opened 2025-04-29
Add constructors for list and union arrays

Alex-PLACET:


include/sparrow/layout/list_layout/list_array.hpp
```diff
@@ -296,6 +297,29 @@ namespace sparrow
...
+            std::optional<METADATA_RANGE> metadata = std::nullopt
+        )
+        {
+            offset_buffer_type list_offsets{std::move(list_offsets_range)};
+            return list_array_impl<BIG>::create_proxy(
+                std::move(flat_values),
+                std::move(list_offsets),
+                std::forward<VB>(validity_input),
+                std::move(name),
+                std::move(metadata)
```
JohanMabille, line 319: Should we forward instead of moving?

Alex-PLACET: Ok fixed

PR #417, opened 2025-05-13
Add iterators for list_value

Alex-PLACET:


include/sparrow/layout/list_layout/list_value.hpp
```diff
@@ -19,9 +19,42 @@
...
 #include "sparrow/types/data_traits.hpp"
+#include "sparrow/utils/iterator.hpp"
 
 namespace sparrow
 {
+
+    class list_value;
+
+    class SPARROW_API list_value_iterator
+        : public iterator_base<list_value_iterator, array_traits::const_reference, std::forward_iterator_tag, array_traits::const_reference>
```
JohanMabille, line 30: This iterator is random access, so you should pass ` std::random_access_iterator_tag`  to the base class.

JohanMabille (changes requested):

JohanMabille (approved):

PR #436, opened 2025-06-02
Rename CONST template argument in run_encoded_array_iterator to avoid…

Alex-PLACET:
… conflict

### DIFF
--- a/include/sparrow/layout/run_end_encoded_layout/run_end_encoded_iterator.hpp
+++ b/include/sparrow/layout/run_end_encoded_layout/run_end_encoded_iterator.hpp
@@ -26,16 +26,16 @@ namespace sparrow
 
     // this iteratas over the **actual** values of the run encoded array
     // Ie nullabes values, not values !!!
-    template <bool SPARROW_CONST>
+    template <bool is_const>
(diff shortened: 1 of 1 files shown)


include/sparrow/layout/run_end_encoded_layout/run_end_encoded_iterator.hpp
```diff
@@ -26,16 +26,16 @@ namespace sparrow
 
     // this iteratas over the **actual** values of the run encoded array
     // Ie nullabes values, not values !!!
-    template <bool CONST>
+    template <bool SPARROW_CONST>
```
JohanMabille, line 29: Nitpicking: Can we use IS_CONST instead of SPARROW_CONST. This latter sounds like a sparrow macro definition for `const`

JohanMabille: Yeah that's actually an argument for switching to lower case `is_const`

Alex-PLACET: done

PR #440, opened 2025-06-03
Add name to record batch and remove nullable flag

Alex-PLACET:


include/sparrow/record_batch.hpp
```diff
@@ -67,7 +67,7 @@ namespace sparrow
                 std::convertible_to<std::ranges::range_value_t<NR>, std::string>
                 and std::same_as<std::ranges::range_value_t<CR>, array>
             )
-        record_batch(NR&& names, CR&& columns);
+        record_batch(std::string_view name, NR&& names, CR&& columns);
```
JohanMabille, line 70: Should we also add (or replace the existing constructor with) a constructor taking a Schema structure?

Alex-PLACET: I don't think so. Maybe add metadata but it does not seems relevant for batches

PR #443, opened 2025-06-04
Add run end encoded integration tests

Alex-PLACET:


src/layout/struct_layout/struct_value.cpp
```diff
@@ -63,21 +63,6 @@ namespace sparrow
...
-    {
-        for (std::size_t i = 0; i < ar.size() - 1; ++i)
-        {
-            std::format_to(ctx.out(), "{}, ", ar[i]);
-        }
-        std::format_to(ctx.out(), "{}", ar[ar.size() - 1]);
-    }
-    return std::format_to(ctx.out(), ">");
-}
-
```
Alex-PLACET, line 80: Moved to cpp to fix issue with clang compiler

PR #446, opened 2025-06-11
Add function to zero null values

Alex-PLACET:


test/test_array.cpp
```diff
@@ -350,7 +350,6 @@ namespace sparrow
                 scalar_value = detail::next_test_value(scalar_value);
             }
 
-            ar.slice(2, 8);
```
Alex-PLACET, line 353: Useless

JohanMabille (approved):

PR #449, opened 2025-06-12
Made struct value iterable

Alex-PLACET:


include/sparrow/layout/struct_layout/struct_value.hpp
```diff
@@ -25,6 +25,40 @@
 
 namespace sparrow
 {
+    class struct_value;
+
+    class SPARROW_API struct_value_iterator : public iterator_base<
```
JohanMabille, line 30: I think this iterator is reimplementing a functor_index_iterator. Defining the following types in `struct_value` should provide the same feature without "code duplication".

```cpp
        using functor_type = detail::layout_value_functor<const struct_value, const_reference>;
        using const_iterator = functor_index_iterator<functor_type>;
```

PR #451, opened 2025-06-16
Add dictionary c data integration tests

Alex-PLACET:


test/c_data_integration/src/comparison.cpp
```diff
@@ -54,27 +54,28 @@ namespace sparrow::c_data_integration
...
+        //     {
+        //         if (std::strcmp(schema->name, schema_from_json->name) != 0)
+        //         {
+        //             differences.push_back(
+        //                 prefix_with_name + " name mismatch: pointer=" + std::string(schema->name)
+        //                 + " vs json=" + std::string(schema_from_json->name)
+        //             );
+        //         }
+        //     }
+        // }
```
JohanMabille, line 78: Wy commenting this?

test/c_data_integration/include/sparrow/c_data_integration/json_parser.hpp
```diff
@@ -40,8 +40,6 @@ namespace sparrow::c_data_integration
     std::vector<sparrow::array>
     get_children_arrays(const nlohmann::json& array, const nlohmann::json& schema, const nlohmann::json& root);
 
-    void read_schema_from_json(const nlohmann::json& data);
```
JohanMabille, line 43: Why dropping this one?

Alex-PLACET: It was unused and useless

test/c_data_integration/src/comparison.cpp
Alex-PLACET: I have to remove it, but the thing is that the dictionary name produced by the C++ implementation doesn't follow the other layout and it's a pain to compare

PR #470, opened 2025-06-26
Pennywise

JohanMabille:
To help reviewers, the implementation comes from [here][https://github.com/xtensor-stack/xtl/blob/master/include/xtl/xhalf_float_impl.hpp) and has been adapted to C++20:
- Replaced comparison operators with `operator<=>`
- Removed operator!= and useless overload of operator== (we now rely on thoses generated by the compiler)
- Added contraints to template parameters of constructor, assignment operator, and comparison operators.
- added format specialization for half_float
- added standard type_traits specializations for half_float


include/sparrow/details/3rdparty/float16_t.hpp
Alex-PLACET, line 1: I think that many methods can be constexpr

Alex-PLACET (approved):

PR #472, opened 2025-06-27
Add release workflow

Alex-PLACET:


JohanMabille (commented):
The main issue is that it does not update the binary version ,and I'm not sure we can easily automate it.

JohanMabille (approved):

PR #479, opened 2025-06-30
Sparrow contracts can now throw

Alex-PLACET:


include/sparrow/buffer/buffer_adaptor.hpp
```diff
@@ -390,8 +390,7 @@ namespace sparrow
 
     template <typename To, BufferReference<To> FromBufferRef>
         requires T_is_const_if_FromBufferRef_is_const<FromBufferRef, To>
-    constexpr buffer_adaptor<To, FromBufferRef>::size_type
-    buffer_adaptor<To, FromBufferRef>::size() const noexcept
+    constexpr buffer_adaptor<To, FromBufferRef>::size_type buffer_adaptor<To, FromBufferRef>::size() const
```
JohanMabille, line 393: You can use a conditional noexcept here:

```cpp
 constexpr buffer_adaptor<To, FromBufferRef>::size_type
buffer_adaptor<To, FromBufferRef>::size() const noexcept(!SPARROW_CONTRACTS_THROW_ON_FAILURE)
```

It might be better to always define SPARROW_CONTRACTS_THROW_ON_FAILURE (that is, set it to 0 if the user does not pass it as a cmake arg), or ro define a shorter alias to it.

include/sparrow/buffer/dynamic_bitset/bitset_iterator.hpp
```diff
@@ -64,9 +64,9 @@ namespace sparrow
         constexpr void increment() noexcept;
         constexpr void decrement() noexcept;
         constexpr void advance(difference_type n) noexcept;
-        [[nodiscard]] constexpr difference_type distance_to(const self_type& rhs) const noexcept;
-        [[nodiscard]] constexpr bool equal(const self_type& rhs) const noexcept;
-        [[nodiscard]] constexpr bool less_than(const self_type& rhs) const noexcept;
+        [[nodiscard]] constexpr difference_type distance_to(const self_type& rhs) const;
+        [[nodiscard]] constexpr bool equal(const self_type& rhs) const;
+        [[nodiscard]] constexpr bool less_than(const self_type& rhs) const;
```
JohanMabille, line 69: Same remarke here

JohanMabille (changes requested):

include/sparrow/buffer/dynamic_bitset/bitset_iterator.hpp
Alex-PLACET: done

Alex-PLACET: done

PR #490, opened 2025-07-04
Turn sparrow exception on for asserte to debug integration tests

Alex-PLACET:


JohanMabille (commented):
I'm not found of incorporating IDE specific stuff into the code base, these should be personal settings).

JohanMabille (approved):

PR #525, opened 2025-08-18
Add documentation

Alex-PLACET:


include/sparrow/timestamp_array.hpp
```diff
@@ -438,6 +438,44 @@ namespace sparrow
          */
         [[nodiscard]] constexpr const_value_iterator value_cend() const;
 
+        /// \cond PRIVATE_DOCS
```
JohanMabille, line 441: Why make this part conditional whereas other private methods are already documented?

PR #535, opened 2025-08-25
Improve struct array and value

Alex-PLACET:


include/sparrow/struct_array.hpp
```diff
@@ -508,6 +558,24 @@ namespace sparrow
...
+        m_children = make_children();
+    }
+
+    template <layout_or_array A>
+    void struct_array::set_child(A&& child, size_t index)
+    {
+        SPARROW_ASSERT_TRUE(child.size() == size());
+        auto [array, schema] = extract_arrow_structures(std::forward<A>(child));
+        get_arrow_proxy().set_child(index, std::move(array), std::move(schema));
+        m_children = make_children();
```
JohanMabille, line 577: Can we avoid rebuilding the whole children list, and just rebuild the one we changed?

include/sparrow/struct_array.hpp
```diff
@@ -242,7 +242,56 @@ namespace sparrow
...
+
+        /**
+         * @brief Adds a child array to the struct.
+         *
+         * @param child The child array to add
+         *
+         * @post Increases the number of children by one
+         */
+        template <layout_or_array A>
+        void add_child(A&& child);
```
JohanMabille, line 269: Should this method accept a range instead (or provide an overload accepting a range) for performance considerations? (i.e. building the children list once for all instead of ubilding it at each add in a loop)

src/layout/struct_value.cpp
```diff
@@ -35,9 +37,31 @@ namespace sparrow
 
     auto struct_value::operator[](size_type i) const -> const_reference
     {
+        SPARROW_ASSERT_TRUE(i < size());
```
JohanMabille, line 40: I agree with copilot, this should be removed, or we should change the definition of https://github.com/man-group/sparrow/blob/main/include/sparrow/utils/contracts.hpp#L172 and set it to 0 by default (and to 1 when building in DEBUG).

JohanMabille: Ah indeed, that should be addressed in a dedicated PR

JohanMabille (approved):

JohanMabille (approved):

PR #540, opened 2025-08-28
Loading sparrow's shared build for emscripten use case

anutosh491:
This should get us started with using sparrow directly with xeus-cpp-lite. We could possibly have an example notebook possibly covering all of sparrow's docs.

image

Once this goes in and there's a release anytime soon, we could just take a xeus-cpp-lite instance and fetch sparrow through 
`%mamba install sparrow` and put it to use !

### DIFF
--- a/include/sparrow/config/config.hpp
+++ b/include/sparrow/config/config.hpp
@@ -63,3 +63,21 @@
 #else
 #    define SPARROW_CONSTEXPR_CLANG constexpr
 #endif
+
+#if defined(__EMSCRIPTEN__) && defined(__CLANG_REPL__)
+#    include <clang/Interpreter/CppInterOp.h>
+
+#    ifndef SPARROW_USE_DATE_POLYFILL
+#        define SPARROW_USE_DATE_POLYFILL 1
+#    endif
+
+#    ifndef HALF_ERRHANDLING_THROWS
+#        define HALF_ERRHANDLING_THROWS 1
+#    endif
+
+static bool _sparrow_loaded = []()
+{
+    Cpp::LoadLibrary("/lib/libsparrow.so", false);
+    return true;
+}();
+#endif
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -1,0 +1,5 @@
+name: sparrow-wasm-host
+channels:
+  - https://repo.prefix.dev/emscripten-forge-dev
+dependencies:
+  - howardhinnant_date


include/sparrow/config/config.hpp
```diff
@@ -63,3 +63,16 @@
...
+#    include <clang/Interpreter/CppInterOp.h>
+
+#    define SPARROW_USE_DATE_POLYFILL 1
+#    define HALF_ERRHANDLING_THROWS 1
+
+static bool _sparrow_loaded = []()
+{
+    Cpp::LoadLibrary("/lib/libsparrow.so", false);
+    return true;
+}();
```
anutosh491, line 77: Explaining a bit of this 

1) I guess config.hpp might be appropriate to address this, hence have added stuff here.

2) We use CppInterOp's LoadLibrary function , that loads a side module (libsparrow.so in this case) over the main module (xcpp.wasm) . The primary `CppInterOp.h` header is always accessible in a xeus-cpp-lite instance, so we can skipping checking its presence.

3) Regarding the two defines, we need those to prevent following errors 

i) SPARROW_USE_DATE_POLYFILL 

Image

ii) HALF_ERRHANDLING_THROWS (We enable HALF_ERRHANDLING_THROWS so that errors in half-precision operations raise C++ exceptions instead of unconditionally calling std::terminate. Also building on top of https://github.com/man-group/sparrow/pull/538)

Image

include/sparrow/config/config.hpp
```diff
@@ -63,3 +63,16 @@
 #else
 #    define SPARROW_CONSTEXPR_CLANG constexpr
 #endif
+
+#if defined(__EMSCRIPTEN__) && defined(__CLANG_REPL__)
+#    include <clang/Interpreter/CppInterOp.h>
+
+#    define SPARROW_USE_DATE_POLYFILL 1
```
Alex-PLACET, line 70: We should handle that at the CMake level.
In the CMakeLists.txt at the root level, you can add a default for the USE_DATE_POLYFILL option to ON when we are on EMSCRIPTEN.
And also add an error message if the USE_DATE_POLYFILL==OFF on EMSCRIPTEN

anutosh491: Ahhh so basically USE_DATE_POLYFILL is already ON by default in CMake, which adds SPARROW_USE_DATE_POLYFILL when building Sparrow itself — but in a xeus-cpp-lite session we don’t consume Sparrow’s CMake targets, so that compile definition isn’t propagated into the REPL, which is why we still need to handle it explicitly there.

So even though our build on emscripten-forge has this USE_DATE_POLYFILL as ON, if we install sparrow, we would still see this 

image

Alex-PLACET: Ha yes I remember. In this case, can you just add a check if it's already defined before you define it ?

anutosh491: > Ha yes I remember. In this case, can you just add a check if it's already defined before you define it ?

Yes, let's do that !

JohanMabille (approved):

PR #542, opened 2025-08-28
Set cpp linter action version to 2.15

Alex-PLACET:


JohanMabille:
Linter change only, no need to wait for CI.

PR #543, opened 2025-08-28
Fixed OSX15 ci

JohanMabille:


JohanMabille:
The fix is not related to exotic architectures, let's merge and not wait longer.

PR #544, opened 2025-08-29
Fixed memory leak in record_batch constructors

JohanMabille:

PR #546, opened 2025-08-29
Removed Upper bound on cmake

JohanMabille:


JohanMabille:
The sanitizer issue is fixed in https://github.com/man-group/sparrow/pull/544

Alex-PLACET (approved):

PR #552, opened 2025-09-02
Homogeneous default allocators

JohanMabille:


Alex-PLACET:
What about the xsimd::aligned_allocator ?

JohanMabille:
This PR does not change the default allocator for buffers: even if `buffer_base` has the xsimd::aligned_allocator by default in its constructors, `buffer`  had was always instantiating an `std::allocator` and calling the buffer_base constructor with it.

Switching to xsimd::aligned_allocator leads to a crash when transferring the ownership of buffer into u8_buffer that I need to investigate. In the meantime, this PR makes the code base homogeneous?

Alex-PLACET (approved):

PR #555, opened 2025-09-05
Copy tracker

JohanMabille:
Failures on Windows should be fixed with #554 
Failures on OSX shoud be fixed with #556

### DIFF
--- a/test/test_record_batch.cpp
+++ b/test/test_record_batch.cpp
@@ -14,8 +14,8 @@
 
 #include <string_view>
 
-#include "sparrow/fixed_width_binary_array.hpp"
 #include "sparrow/debug/copy_tracker.hpp"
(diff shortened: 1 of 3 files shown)


test/test_record_batch.cpp
```diff
@@ -199,9 +200,16 @@ namespace sparrow
 
         TEST_CASE("copy semantic")
         {
+#ifdef SPARROW_TRACK_COPIES
```
Alex-PLACET, line 203: I think we should always track the copies in the tests

Alex-PLACET (approved):

JohanMabille:
Yeah, I need to fix the conflicts first.

Alex-PLACET (approved):

PR #573, opened 2025-10-03
Remove osx13 from workflow as it will be deprecated soon

Alex-PLACET:


JohanMabille:
CI change only, let's merge.

PR #594, opened 2025-10-20
Add uuid array

Alex-PLACET:


JohanMabille (commented):
Instead of extending `fixed_width_binary_array` "by the bottom", could we extend it "by the top"? `uuid_array` simply adds static metadata to a `fixed_width_binary_array`, so I would find it simpler to have something like:

```cpp
template <std::ranges::sized_range T, class CR, class Ext>
class fixed_width_binary_array_impl
    : public mutable_array_bitmap_base<fixed_width_binary_array_impl<T, CR, Ext>>
    , public Ext
{
public:

    // as before ..

private:

    arrow_proxy create_proxy_impl(...)
    {
        arrow_proxy = // as before
        Ext::init(proxy);
        return proxy;
    }  
};
```

And then

```cpp
struct empty_extension
{
protected:
    void init(arrow_proxy&) {}
};

struct uuid_extension
{
public:
    static constexpr std::string_view EXTENSION_NAME = "arrow.uuid";
protected:
    void init(arrow_proxy&) {}
};

using fixed_width_binary_array = fixed_width_binary_array_impl<
    fixed_width_binary_traits::value_type,
    fixed_width_binary_traits::const_reference,
    empty_extension
>;

using uid_array = fixed_width_binary_array_impl<
    fixed_width_binary_traits::value_type,
    fixed_width_binary_traits::const_reference,
    uid_extension
>;
```

(Of course the extension and uuid_array types should be defined in their own header)

I don't know if we can use this pattern for other extension types, but it seems quite relevant here.

include/sparrow/uuid_array.hpp
```diff
@@ -0,0 +1,111 @@
...
+    struct uuid_extension
+    {
+    public:
+
+        static constexpr size_t UUID_SIZE = 16;
+        static constexpr std::string_view EXTENSION_NAME = "arrow.uuid";
+
+    protected:
+
+        static void init(arrow_proxy& proxy)
```
JohanMabille: Same here, no need to make this method static.

PR #597, opened 2025-10-22
Record batch array by reference

Alex-PLACET:


include/sparrow/record_batch.hpp
```diff
@@ -493,13 +562,22 @@ namespace sparrow
         SPARROW_API void check_consistency() const;
 
         using metadata_type = std::vector<metadata_pair>;
+        using array_storage_type = std::variant<array, std::reference_wrapper<array>>;
+
         std::optional<name_type> m_name = std::nullopt;          ///< Optional name of the record batch
         std::optional<metadata_type> m_metadata = std::nullopt;  ///< Optional metadata for the record batch
         std::vector<name_type> m_name_list;                      ///< Ordered list of column names
-        std::vector<array> m_array_list;                         ///< Ordered list of column arrays
+        std::vector<array_storage_type> m_array_list;            ///< Ordered list of column arrays (owned or
```
JohanMabille: Do we want to be able to mix column references and values? From the implementaiton of the constructor, it seems the contrary. Therefore it may be more efficient to have a `std::variant<std::vector<array>, std::vector<std::reference_wrapper<array>>>`

include/sparrow/record_batch.hpp
```diff
@@ -539,8 +617,19 @@ namespace sparrow
...
+        {
+            if constexpr (std::is_lvalue_reference_v<CR>)
+            {
+                m_array_list.emplace_back(col);
+            }
+            else
+            {
+                m_array_list.emplace_back(std::move(col));
+            }
+        }
```
JohanMabille: If we store a variant of vector instead of a vector of variant, I think this method can be simplified with forwarding (it might require smoe help regarding intermediate types).

include/sparrow/record_batch.hpp
JohanMabille: From in person conversation, this is actually a use-case we must support.

JohanMabille (approved):

PR #601, opened 2025-10-29
Fix metadata copy in schema

Alex-PLACET:


include/sparrow/utils/metadata.hpp
```diff
@@ -287,6 +287,37 @@ namespace sparrow
...
+         * Compares two key_value_view objects for equality by checking if they contain
+         * the same key-value pairs in the same order.
+         *
+         * @param lhs First view to compare
+         * @param rhs Second view to compare
+         * @return true if both views contain identical key-value pairs in the same order
+         *
+         * @post Returns true iff both views have the same size and all pairs are equal
+         */
+        friend bool operator==(const key_value_view& lhs, const key_value_view& rhs)
```
JohanMabille: I think this one can be implemented with `std::tanges::equal`. Besides, since it does not use any private member, the operator can be declared as a free function out of the class.

JohanMabille (approved):

PR #603, opened 2025-10-29
Add `sparrow-json-reader` to conan

Hind-M:


Hind-M:
> I forget to update this conan recipe. Can you take into account the modification I made on conan center ? https://github.com/conan-io/conan-center-index/blob/master/recipes/sparrow/all/conanfile.py

Done. I think we should also change the conan center recipe to set `transitive_headers` in `nlohmann_json` as well:
```
  def requirements(self):
        if self._uses_date_polyfill:
            self.requires("date/3.0.3", transitive_headers=True)
        if self.options.export_json_reader:
            self.requires("nlohmann_json/3.12.0", transitive_headers=True)
```

PR #611, opened 2025-11-17
Array registry for dispatch and factory

Alex-PLACET:


include/sparrow/layout/array_registry.hpp
```diff
@@ -0,0 +1,442 @@
...
+         *
+         * This overload allows for more complex extension detection logic beyond
+         * simple metadata name matching.
+         *
+         * @param base_type The underlying base data_type
+         * @param predicate Custom function to check if proxy is this extension
+         * @param factory Factory function to create the array
+         */
+        SPARROW_API void
+        register_extension_with_predicate(data_type base_type, extension_predicate predicate, factory_func factory);
```
JohanMabille: This should be named `register_extension`  too, the compiler can figure out the right overload at the call site.

include/sparrow/layout/array_registry.hpp
```diff
@@ -0,0 +1,442 @@
...
+                }
+                else
+                {
+                    return try_dispatch_recursive<F, I + 1>(std::forward<F>(func), ar, dt);
+                }
+            }
+            else
+            {
+                throw std::invalid_argument("array type not supported");
+            }
```
JohanMabille: I think we could get rid of the recursion with a "table", something like

```cpp
template <class F>
struct invoker
{
    template <class I>
    static auto run(F& f, const array_wrapper& ar)
    {
        return f(unwrap_array<array_type_t<I>>(ar));
    }
};

template <class F, std::size_t... I>
constexpr auto make_invoker_table(F& f, std::index_sequence<I...>)
{
    return std::array<invoker_type, all_data_types.size()>{&invoker<F>::run<I>...};
}

template <class F>
visit_result_t<F> dispatch(F&& f, const array_wrapper& ar)
{
    const auto dt = ar.data_type();
    if (dt >= all_data_types.size())
    {
        throw ...;
    }
    return make_invoker_table(f, std::make_index_sequence(all_data_types.size())[dt](ar);
}
```

This also avoids checking boundary condition on dt at each step.

include/sparrow/layout/array_registry.hpp
```diff
@@ -0,0 +1,521 @@
...
+                return dispatch_for_type<F, DT>(std::forward<F>(func), ar);
+            }
+        };
+
+        template <class F>
+        static consteval auto make_dispatch_table()
+        {
+            using result_t = visit_result_t<F>;
+            using invoker_t = result_t (*)(F&&, const array_wrapper&);
+            constexpr std::size_t table_size = std::numeric_limits<std::underlying_type_t<data_type>>::max()
```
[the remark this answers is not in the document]
Alex-PLACET: I fixed the issue by removing the enum values for data_type. Now I'm using all_data_types.size()

JohanMabille (approved):

PR #626, opened 2025-12-15
Adjust instantiation order to match declaration

eddelbuettel:
In another project with `sparrow` vendore, the compiler complains about the instantation order.  This short PR ensure it matches the declaration 

https://github.com/man-group/sparrow/blob/b1b8ae12391bce18bdecb9261c0ea9de7dd3ae30/src/arrow_interface/arrow_array_schema_proxy.cpp#L213-L215

so that the warning goes away.


Alex-PLACET (approved):
Thank you for your PR. :+1: 
I will add checks for members initialization order in a next PR.

Alex-PLACET:
@eddelbuettel What is your compiler's version ?

Alex-PLACET:
We currently compile with these options: https://github.com/man-group/sparrow/blob/main/cmake/compile_options.cmake
And we use gcc from 11 to 14 and clang from 17 to 20.
I will investigate that

PR #637, opened 2026-01-07
Migrate to emscripten 4-x

anutosh491:
We're using sparrow on the 4-x branch of emscripten-forge. We should reflect the same changes on master.


anutosh491:
cc @Alex-PLACET should be all the changes we need for the migration, not sure if the first 4 failing jobs are related !

PR #638, opened 2026-01-07
Enable more tests for the emscripten build

anutosh491:
Most of the symbol resolution errors we had last time with these tests are now resolved after the migration to 4-x.


anutosh491:
cc @Alex-PLACET for review

PR #639, opened 2026-01-09
Create bit vector

Alex-PLACET:


include/sparrow/buffer/bit_vector/bit_vector_base.hpp
Alex-PLACET: Comes from the dynamic_bitset_base without the null_count

PR #642, opened 2026-01-13
Add libpopcnt

Alex-PLACET:


JohanMabille (changes requested):
This will prevent to build the feedstock until libpopcnt is available on conda-forge (or we will have to patch the feedstock). Vendoring the file under `details/3rdparty` would be easier.

PR #645, opened 2026-01-20
Remove 'push' event condition in manylinux CI

Alex-PLACET:


JohanMabille (changes requested):
I don't think we should remove this condition. When opening a PR, this would trigger builds on both main and the PR branch.

PR #670, opened 2026-04-01
Run end encoded mutability

Alex-PLACET:


include/sparrow/array_api.hpp
```diff
@@ -430,13 +437,6 @@ namespace sparrow
 
     private:
 
-        /**
-         * Constructs an array from an arrow proxy.
-         *
-         * @param proxy The arrow proxy to construct from.
-         */
-        SPARROW_API array(arrow_proxy&& proxy);
-
```
Alex-PLACET, line 439: Moved to public

JohanMabille: Why?

include/sparrow/run_end_encoded_array.hpp
```diff
@@ -18,13 +18,51 @@
...
 #include "sparrow/layout/array_access.hpp"
 #include "sparrow/layout/array_wrapper.hpp"
-#include "sparrow/layout/run_end_encoded_iterator.hpp"
 #include "sparrow/utils/memory.hpp"
 
 namespace sparrow
 {
     class run_end_encoded_array;
 
+    class run_end_encoded_reference : public array_traits::value_type
```
JohanMabille: Why do we need this inheritance? This is usually dangerous to mix public inheritance and value semantic as it can lead to sliced assigments.

include/sparrow/run_end_encoded_array.hpp
```diff
@@ -364,11 +557,11 @@ namespace sparrow
         ArrowSchema** child_schemas = new ArrowSchema*[n_children];
         ArrowArray** child_arrays = new ArrowArray*[n_children];
 
-        child_schemas[0] = new ArrowSchema(std::move(acc_length_schema));
-        child_schemas[1] = new ArrowSchema(std::move(encoded_values_schema));
+        child_schemas[0] = new ArrowSchema(acc_length_schema);
+        child_schemas[1] = new ArrowSchema(encoded_values_schema);
```
JohanMabille: The extracted schemas should be moved.

include/sparrow/run_end_encoded_array.hpp
```diff
@@ -364,11 +557,11 @@ namespace sparrow
...
 
-        child_arrays[0] = new ArrowArray(std::move(acc_length_array));
-        child_arrays[1] = new ArrowArray(std::move(encoded_values_array));
+        child_arrays[0] = new ArrowArray(acc_length_array);
+        child_arrays[1] = new ArrowArray(encoded_values_array);
```
JohanMabille: Same here, the arrays should be moved instead of being passed by reference.

src/run_end_encoded_array.cpp
```diff
@@ -14,15 +14,258 @@
...
 #include "sparrow/layout/array_helper.hpp"
 #include "sparrow/layout/array_registry.hpp"
 #include "sparrow/primitive_array.hpp"
-#include "sparrow/utils/repeat_container.hpp"
 
 namespace sparrow
 {
+    template <class T>
+    concept usable_array = (mpl::is_type_instance_of_v<T, primitive_array>
+                            || mpl::is_type_instance_of_v<T, primitive_array_impl>)
```
JohanMabille: - `primitive_array`  is an alias on `primitive_array_impl`, so testing only this last one should be enough.
- This concept could be moved to the anonymous namespace below.

JohanMabille (changes requested):

include/sparrow/array_api.hpp
Alex-PLACET: I need it in the [rebind_children_from_proxy](https://github.com/man-group/sparrow/pull/670/changes#diff-7aa5f81ddea75ca00c73393af498edaa4deba08858399b6d842da8d989fbf607R644)

src/run_end_encoded_array.cpp
Alex-PLACET: fixed

include/sparrow/run_end_encoded_array.hpp
Alex-PLACET: fixed

include/sparrow/run_end_encoded_array.hpp
Alex-PLACET: fixed

include/sparrow/run_end_encoded_array.hpp
Alex-PLACET: fixed

JohanMabille (approved):

PR #676, opened 2026-06-16
fix(arrow_interface): replace brace-init array-new with explicit assignment for MSVC compat

### LINKED ISSUE #675: map_array::create_proxy_impl fails to compile on MSVC 14.51
Environment
  - sparrow version: 2.4.0
  - Compiler: MSVC cl.exe 19.51 / toolset 14.51.36362 (VS 2022 17.14, the toolset on the current GitHub Actions windows-latest / Windows Server 2025 image)
  - Flags: /std:c++20 /permissive- /EHsc
  - GCC and AppleClang (also /std:c++20) compile the same code without error; this reproduces only on the newer MSVC.

## Summary

Including <sparrow/map_array.hpp> (transitively, via <sparrow/record_batch.hpp>) and instantiating map_array fails to compile under MSVC 14.51. It compiled cleanly on the previous windows-latest MSVC toolset, so this surfaced purely as a toolchain bump rather than any change on our ArcticDB CI— the newer MSVC's stricter overload resolution rejects a conversion the older one accepted.

Example of a failing run [here](https://github.com/man-group/ArcticDB/actions/runs/27551106425/job/81437338937).

### Compiler output

``` bash
  sparrow/map_array.hpp(589): error C2672: 'make_arrow_array': no matching overloaded function found
  sparrow/arrow_interface/arrow_array.hpp(63): note: could be 'ArrowArray sparrow::make_arrow_array(int64_t,int64_t,int64_t,B,ArrowArray **,const CHILDREN_OWNERSHIP &,ArrowArray *,bool)'
  sparrow/map_array.hpp(594): note: 'initializing': cannot convert from 'ArrowArray *(*)[1]' to 'ArrowArray **'
  sparrow/map_array.hpp(594): note: Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or parenthesized function-style cast
  sparrow/map_array.hpp(589): note: 'make_arrow_array': function declaration must be available as none of the arguments depend on a template parameter
  sparrow/map_array.hpp(359): note: This diagnostic occurred in the compiler generated function
      'sparrow::arrow_proxy sparrow::map_array::create_proxy_impl(...)'
```

Alex-PLACET:
Fix MSVC build failure caused by `new T*[N]{...}` brace-initialized array-new expressions.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -39,14 +39,12 @@ jobs:
         - { name: Debug }
         - { name: Release }
         build-system:
-        - { name: "Visual Studio 17 2022" }
-        - { name: "Visual Studio 18 2026" }
-        - { name: "Ninja" }
-        exclude:
+        - {name: "Ninja"}
+        include:
         - runs-on: windows-latest
-          build-system: { name: "Visual Studio 18 2026" }
+          build-system: {name: "Visual Studio 17 2022"}
         - runs-on: windows-2025-vs2026
-          build-system: { name: "Visual Studio 17 2022" }
+          build-system: {name: "Visual Studio 18 2026"}
 
     steps:
     - name: Setup MSVC (only for non-VS buildsystems)
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -42,9 +42,9 @@ jobs:
         - {name: "Ninja"}
         include:
         - runs-on: windows-latest
-          build-system: {name: "Visual Studio 17 2022"}
+          build-system: {name: "Visual Studio 17 2022", vs-version-range: "[17.0,18.0)"}
         - runs-on: windows-2025-vs2026
-          build-system: {name: "Visual Studio 18 2026"}
+          build-system: {name: "Visual Studio 18 2026", vs-version-range: "[18.0,19.0)"}
 
     steps:
     - name: Setup MSVC (only for non-VS buildsystems)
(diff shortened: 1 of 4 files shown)


ci/docker/integration.dockerfile
```diff
@@ -34,7 +34,7 @@ ENV BUILD_DOCS_CPP=OFF
 
 RUN apt update
 
-RUN apt install build-essential git gcc -y
+RUN apt install build-essential git gcc -y
```
JohanMabille: This can be reverted.

.github/workflows/integration_tests.yaml
```diff
@@ -4,14 +4,14 @@ on:
   workflow_dispatch:
   pull_request:
   push:
-    branches: [main]
+    branches: [ main ]
```
JohanMabille: Is that a new formatting requirement? Otherwise I think it should be reverted.

.github/workflows/windows.yml
```diff
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
   pull_request:
   push:
-    branches: [main]
+    branches: [ main ]
```
JohanMabille: Same question here.

.github/workflows/windows.yml
```diff
@@ -17,11 +17,11 @@ jobs:
...
       matrix:
-        runs-on: [windows-latest]
+        runs-on: [ windows-2022, windows-2025-vs2026 ]
         toolchain:
-        - {compiler: msvc, date-polyfill: 'ON', shared: 'ON'}
-        - {compiler: msvc, date-polyfill: 'ON', shared: 'OFF'}
-        - {compiler: msvc, date-polyfill: 'OFF', shared: 'ON'}
+        - { compiler: msvc, date-polyfill: 'ON', shared: 'ON' }
+        - { compiler: msvc, date-polyfill: 'ON', shared: 'OFF' }
+        - { compiler: msvc, date-polyfill: 'OFF', shared: 'ON' }
```
JohanMabille: Same here.

JohanMabille (approved):
LGTM when the formatting issues have been addressed.

PR #678, opened 2026-08-07
Fix build with doctest 2.5.x

rosset:
In doctest 2.5.x, XmlWriter is defined in the doctest::detail namespace rather than the doctest namespace directly.

This patch qualifies the type correctly as doctest::detail::XmlWriter to resolve the build issue.


Alex-PLACET:
It as been fixed here: https://github.com/sparrow-org/sparrow

